PR #6, opened 2015-11-25
Add guidelines stub

phadej:


README.md
```diff
@@ -25,3 +25,55 @@ wc */src/**/*.hs
...
+[Example
+`stylish-haskell.yaml`](https://github.com/futurice/haskell-servant-status/blob/master/.stylish-haskell.yaml).
+
+Packages have to be buildable with `-Wall -Werror` with some of the supported
+GHCs (currently usually *7.10.2*).
+
+### GHC Extensions
+
+See [ghccaniuse](http://damianfral.github.io/ghcaniuse/). Don't use extensions,
+which aren't supported by three or four GHC releases. i.e. using `DataKinds`
```
zudov: Shouldn't it apply only to 'libraries'? e.g. "In case of libraries adhere to `[Three release policy](link to three-release-policy section)`.

phadej: True. `PartialTypeSignatures` is probably a good example of something one might want to use somewhere. Will change.

PR #305, opened 2017-04-05
Allow 00 based international syntax

mixman:


sms-proxy/package.yaml
```diff
@@ -17,17 +17,18 @@ library:
...
     - futurice-prelude
     - futurice-servant
-    - servant-server
     - http-client
     - http-client-tls
     - http-types
-    - aeson
-    - aeson-compat
-    - aeson-extra
+    - regex-compat
```
phadej: ei

sms-proxy/src/Futurice/App/SmsProxy/Logic.hs
```diff
@@ -12,23 +12,27 @@ import Futurice.App.SmsProxy.Ctx
...
 
+import Text.Regex (subRegex, mkRegex)
+import qualified Data.Text as T
+
 sendSms :: (MonadIO m, MonadLog m) => Ctx -> Req -> m Res
 sendSms ctx req = do
-    -- Log "metadata"
     logInfo_ $ "Sending message to " <> req ^. reqTo
 
+    let number = T.pack $ subRegex (mkRegex "^00") (T.unpack $ req ^. reqTo) "+"
```
phadej: use either http://hackage.haskell.org/package/regex-applicative-text-0.1.0.1/docs/Text-Regex-Applicative-Text.html
or `T.isPrefixOf`, `T.drop 2` and `<>`.

phadej: FWIW, grep the repo, there are uses of `T.R.A.Text`.

sms-proxy/package.yaml
mixman: Routine sort needs sort -i; didn't even look back at the results.

sms-proxy/src/Futurice/App/SmsProxy/Logic.hs
mixman: What's the benefit of the different RE package? Better suited for the types used?

phadej: no syntax errors in re syntax. And no reason to use many libraries for the same thing.

phadej:
I'll deploy this shortly (will build `planmill-proxy` anyway too soon).

PR #342, opened 2017-05-31
added employee department and office

ArktinenSieni:
Added Employee's office name and department name. Name fields were used since personio-api didn't seem to have support for getting either office or department by their id. If required, names can easily be replaced with ids.


phadej:
names are good enough.

phadej:
In https://github.com/futurice/haskell-mega-repo/pull/344

PR #346, opened 2017-06-01
Smileys refactor

phadej:


smileys-app/schema.psql
```diff
@@ -1,12 +1,13 @@
 CREATE SCHEMA smileys;
 
 CREATE TABLE smileys.trail (
-  id SERIAL PRIMARY KEY,
   entries JSON NOT NULL,
   username TEXT NOT NULL,
   smiley SMALLINT NOT NULL,
   day DATE NOT NULL,
-  constraint one_a_day UNIQUE (username, day)
+  PRIMARY KEY (username, day)  -- this is natural primary key
```
mixman: What's the benefit of a natural PK here?

smileys-app/schema.psql
```diff
@@ -1,12 +1,13 @@
...
 );
 
-CREATE INDEX ON smileys.trail (day);
+-- primary key creates an index,
+-- but let's have this way around too.
+CREATE INDEX ON smileys.trail (day, username);
```
mixman: Does this also allow efficient lookups on day alone (start-date, end-date -range lookups)?

smileys-app/src/Futurice/App/Smileys/Logic.hs
```diff
@@ -14,43 +14,43 @@ import qualified FUM
...
-    let today = utctDay now
-    withResource (ctxPostgresPool ctx) $ \conn -> do
-        res <- q conn (fromMaybe today start) (fromMaybe today end)
-        return $ res
-        where
-          q :: (MonadIO m, MonadBaseControl IO m) => Postgres.Connection -> Day -> Day -> m [Smileys]
-          q conn s e = liftIO $ Postgres.query conn
-                "SELECT entries, username, smiley, day FROM smileys.trail WHERE day >= ? AND day <= ?"
-                (s, e)
+getSmileys ctx mfum start end =
```
mixman: I'll move this to a getSmileysForUser as getSmileys is needed also as-is.

smileys-app/schema.psql
phadej: yes.

smileys-app/src/Futurice/App/Smileys/Logic.hs
phadej: I'd rather have three
- `getSmileys`
- `getSmileysForAll` (without SSO User)
- `getSmileysForUser` (Where FUM.UserName` is a `Capture`

phadej: Or last two can be the same, if `FUM.UserName` is a `QueryParam` (return all if `Nothing`).

smileys-app/schema.psql
phadej: > A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns. 

etc. https://www.postgresql.org/docs/9.6/static/indexes-multicolumn.html

smileys-app/schema.psql
phadej: No need for the artificial one.

mixman: They are two ways of achieving the same goal; both have pro/cons depending on the situation. Was wondering why in this case a NPK.

phadej:
travis failure was the timeout.

PR #354, opened 2017-06-06
validate email

ArktinenSieni:
- Validates email

One question which arose from line 408 in Types.hs: should array give different error if array length is greater than 0. For example emailInvalidFormat or emailInvalidType.

### DIFF
--- a/personio-client/src/Personio/Types.hs
+++ b/personio-client/src/Personio/Types.hs
@@ -383,7 +383,9 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
 
     validate :: HashMap Text Attribute -> Parser [ValidationMessage]
     validate obj = execWriterT $ sequenceA_
-        [ validateGithub
+        [
+            validateGithub,
+            emailMissing
         ]
       where
         validateGithub :: WriterT [ValidationMessage] Parser ()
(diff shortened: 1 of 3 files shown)


personio-client/src/Personio/Types.hs
```diff
@@ -383,7 +383,9 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
 
     validate :: HashMap Text Attribute -> Parser [ValidationMessage]
     validate obj = execWriterT $ sequenceA_
-        [ validateGithub
+        [
```
phadej: we use front comma formatting:

```haskell
[ validateBithub
, emailMissing
]
```

personio-client/src/Personio/Types.hs
```diff
@@ -396,4 +398,19 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
             regexp :: RE' Text
             regexp = string "https://github.com/" *> (T.pack <$> some anySym)
 
+        anyText :: Text -> Maybe Text
+        anyText val = match (T.pack <$> some anySym :: RE' Text) val
```
phadej: maybe name it `isSomeText` (`anyText` would be `Just`, i.e. accept *anything* :) )

personio-client/src/Personio/Types.hs
```diff
@@ -396,4 +398,19 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+
+        emailMissing :: WriterT [ValidationMessage] Parser ()
+        emailMissing = do
+            email <- lift (parseAttribute obj "email")
+            case email of
+                (Array _) -> tell [EmailMissing] -- | Assumes that if email is given as array, it is missing
+                _         -> checkMail (T.pack (show email))
```
phadej: `T.pack (show`, that's probably not what you want to do. You probably wnt to match on `String` constructor of [`Value`](http://hackage.haskell.org/package/aeson-1.2.0.0/docs/Data-Aeson-Types.html#t:Value)

ArktinenSieni: I followed something similar to this: https://en.wikibooks.org/wiki/Haskell/Pattern_matching like this: 
`(String v) -> checkMail v`
And it gave error 
```
Pattern match(es) are non-exhaustive
In a case alternative:
    Patterns not matched:
        (Object _)
        (Number _)
        (Bool _)
        Null
```
I think I misunderstood? Should I instead just give isSomeText a String which uses T.pack anyways?

phadej: add a "default" case, `v -> typeMismatch "Email" v` http://hackage.haskell.org/package/aeson-1.2.0.0/docs/Data-Aeson-Types.html#v:typeMismatch, there's also an example.

ArktinenSieni:
Cherry-picked to other PR

PR #360, opened 2017-06-20
Iban check refactor

phadej:


phadej:
ping @ArktinenSieni

personio-client/src/Personio/Types.hs
```diff
@@ -483,30 +487,47 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+    -- '2' -> (\x -> x * 10 + 2)
+    -- 'a' -> (\x -> x * 100 + 10)
+    --
+    -- etc.
+    --
+    -- Note: we need to fold from the left!
+    --
+    toDigit :: Char -> Maybe (Integer -> Integer)
+    toDigit c
+        | '0' <= c && c <= '9' = Just $ \n -> n * 10 + fromIntegral (ord c - ord '0')
```
ArktinenSieni: What value does n receive? It is index number from traverse?

personio-client/src/Personio/Types.hs
```diff
@@ -483,30 +487,47 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+        | 'A' <= c && c <= 'Z' = Just $ \n -> n * 100 + fromIntegral (ord c - ord 'A' + 10)
+        | otherwise            = Nothing
+
+    -- takes spaceless and re-arranged string
+    validate :: String -> Bool
+    validate xs = isJust $ do
+        ds <- traverse toDigit xs  -- replace characters
```
ArktinenSieni: Traverse's documentation said it evaluates actions and collects results. I do not understand this collection process.

personio-client/src/Personio/Types.hs
```diff
@@ -483,30 +487,47 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+        guard (15 <= length ds && length ds <= 31) -- length check
```
ArktinenSieni: - Was travel length-check meant to be done to IBAN before replacing characters or after it?
- How does this function behave when guard returns Nothing? It just breaks the "do"?

personio-client/src/Personio/Types.hs
phadej: Refer to `Traversable` chapter (21) in the haskellbook.

That chapter starts with using `Maybe` as an `Applicative`, as we do here, you should be able to quickly get what `traverse` does here (as opposed to `catMaybes` / `mapMaybe` would do).

personio-client/src/Personio/Types.hs
phadej: note, `length xs == length ds`. `traverse` doesn't change the *shape* of the containers.

I.e. if `xs` is `"A3"`, then `ds` is `[ \x -> x * 100 + 10, \x -> x * 10 + 3 ]`. Length doesn't change.

---

About: `guard` check it's definition, inline the `Maybe`s `>>` and check what happens!

```haskell
(>>) :: Maybe a -> Maybe b -> Maybe b
Nothing >> _ = Nothing
Just _ >> x = x
```

personio-client/src/Personio/Types.hs
ArktinenSieni: Probably I don't get this since I do not understand folding too well yet?

personio-client/src/Personio/Types.hs
ArktinenSieni: Understood what it does, but not yet how it does :D

personio-client/src/Personio/Types.hs
```diff
@@ -483,30 +487,47 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+        -- Interpret the string as a decimal integer
+        -- and compute the remainder of that number on division by 97
+        let checksum = foldl' (&) 0 ds
```
ArktinenSieni: Folding is still beyond me.

phadej: Oh, that one is a bit tricky indeed.

```haskell
  fold' (&) 0 ds
= foldl' (\x f -> f x) 0 ds
```
---

`foldl'` folds from the left:

```haskell
  foldl' f i [a, b, c]
= ((i `f` a) `f` b) `f` c
```

---

if we have three element `ds`:

```haskell
  foldl' (&) 0 [f, g, h]
= 0 & a & b & c
```

let's say we started with `2A3` string, then `ds` is

```haskell
[ \x -> x * 10 + 2, \x -> x * 100 + 10, \x -> x * 10 + 3 ]
```

Combining the previous two we get:

```haskell
  foldl' (&) 0 ds
= 0 & (\x -> x * 10 + 2) & (\x -> x * 100 + 10) & (\x -> x * 10 + 3)
```

we can reverse the chain of `&` -> `$`

```haskell
= 0 & (\x -> x * 10 + 2) & (\x -> x * 100 + 10) & (\x -> x * 10 + 3)
= (\x -> x * 10 + 3) $ (\x -> x * 100 + 10) $ (\x -> x * 10 + 2) $ 0
= (\x -> x * 10 + 3) ((\x -> x * 100 + 10) ((\x -> x * 10 + 2) 0))
```

and then do the reductions:

```haskell
= (\x -> x * 10 + 3) ((\x -> x * 100 + 10) ((\x -> x * 10 + 2) 0))
= (\x -> x * 10 + 3) ((\x -> x * 100 + 10) (0 * 10 + 2))
= (\x -> x * 10 + 3) ((\x -> x * 100 + 10) 2)
= (\x -> x * 10 + 3) (2 * 100 + 10))
= (\x -> x * 10 + 3) 210)
= 210 * 10 + 3
= 2103
```

which is what we get from `2A3` -> "2 10 3"

PR #367, opened 2017-07-03
filter swagger logs on client-side

ArktinenSieni:
Cleaning proxy-app's log from swagger-entries. This could also be done altering "acceslog"-table (in /proxy-mgmt-app/schema.psql) and adding a condition to column "endpoint" to filter swagger-entries.

### DIFF
--- a/proxy-app/src/Futurice/App/Proxy.hs
+++ b/proxy-app/src/Futurice/App/Proxy.hs
@@ -28,7 +28,9 @@ import Servant.Common.Req              (Req (headers))
 import Servant.Proxy
 import Servant.Excel
 import System.IO                       (hPutStrLn, stderr)
+import Text.Regex.Applicative.Text     (RE', anySym, match, string)
 
+import qualified Data.Text                  as T
 import qualified Database.PostgreSQL.Simple as Postgres
 import qualified FUM
 import qualified Futurice.GitHub            as GH (SomeRequest, SomeResponse)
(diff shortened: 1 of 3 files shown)


proxy-app/src/Futurice/App/Proxy.hs
```diff
@@ -205,7 +207,17 @@ checkCreds ctx req u p = withResource (ctxPostgresPool ctx) $ \conn -> do
             pure False
         _ : _ -> do
             let endpoint = decodeLatin1 $ rawPathInfo req
-            _ <- Postgres.execute conn
-                "insert into proxyapp.accesslog (username, endpoint) values (?, ?);"
-                (u', endpoint)
+            _ <- logAccess conn u' endpoint
```
mixman: Using an exclusion rule, then `_ <- return (match isNotSwaggerReg endpoint) >>= (\_ -> Postgres...) might be an option.

phadej: For me, explicit use of `>>=` feels almost always fishy, writing in `do` notation is much clearer

I guess the intention of @mixman comment was:
```haskell
do
    _ <- return (match isNotSwaggerReg endpoint)
    Postgres.execute ...
``` 

I'm curious, why you use `>>=`?

If you try out things in GHCi, then you can use `do` there as well!

```haskell
Prelude> do { x <- Just 1; y <- Just 2; return $ 1 + 2 }
Just 3
```

*Note:* I recall myself using something like:
```haskell
thingWhichReturnsEither >>= either throwError pure
```
but really I should make a combinator and use it. Something like https://www.stackage.org/haddock/lts-8.21/either-4.4.1.1/Control-Monad-Trans-Either.html#v:hoistEither but more general.

proxy-app/src/Futurice/App/Proxy.hs
```diff
@@ -210,14 +211,13 @@ checkCreds ctx req u p = withResource (ctxPostgresPool ctx) $ \conn -> do
...
-      where
-        swaggerReg :: RE' Text
-        swaggerReg = string "/swagger" *> (T.pack <$> many anySym)
+        when (isNothing $ match isSwaggerReg endp) $ void $
+            Postgres.execute conn
+            "insert into proxyapp.accesslog (username, endpoint) values (?, ?);"
+            (user, endp)
+
+    isSwaggerReg :: RE' Text
+    isSwaggerReg = string "/swagger.json" <|> string "/swagger-ui" *> (T.pack <$> many anySym)
```
ArktinenSieni: How could I make a "isNotSwaggerReg" in similar fashion? What would the syntax be then?

phadej: Unfortunately there aren't complement operation in `regex-applicative` (though regular expressions are closed under complement). If possible change "if matches" to "if doesn't match"

PR #368, opened 2017-07-10
parse 'status' and 'HR number', validations check

mixman:


personio-client/src/Personio/EmployeeStatus.hs
```diff
@@ -0,0 +1,57 @@
...
+import Data.Aeson.Compat (Value (String), withText)
+import Futurice.Generics
+import Futurice.Prelude
+import Prelude ()
+import Data.Swagger               (NamedSchema (..))
+
+import qualified Data.Map  as Map
+import qualified Data.Text as T
+
+data Status
```
phadej: We have to document how this is related to the `hireDate` and `contractEndDate`, and what to do when they don't agree (e.g. `status = Active` but `contractEndDate` passed OR `status = Inactive` but `hireDate` passed etc).

phadej:
General TODO: refactor to use `Futurice.Enum` like https://github.com/futurice/haskell-mega-repo/blob/a0af7cd125c40ebc65f3ae10b646f38032859784/fum-carbon-app/src/Futurice/App/FUM/Types/Status.hs

This cannot be done right now, as `Futurice.Enum` is WIP in `fum-carbon-app` (will be moved to `futurice-prelude`, eventually, but later defining this kind of enum types will be simpler

personio-client/package.yaml
```diff
@@ -17,8 +17,10 @@ ghc-options: -Wall
 dependencies:
   - aeson
   - aeson-compat
+  - ansi-pretty
```
phadej: ?

mixman: Would be useful to be available from Futurice.Prelude

personio-client/src/Personio/Types.hs
```diff
@@ -196,6 +197,7 @@ instance FromJSON SupervisorId where
 newtype NamedAttribute = NamedAttribute { getName :: Maybe Text }
 
 instance FromJSON NamedAttribute where
+    parseJSON (Null) = pure (NamedAttribute Nothing)
     parseJSON v = case v of
```
phadej: would be better to move into `case` below, so we don't pattern match on the argument twice (though GHC can optimise those, yet still).

personio-client/package.yaml
phadej: I mean, I don't see any uses of `ansi-pretty` in the source

personio-client/src/Personio/EmployeeStatus.hs
```diff
@@ -0,0 +1,57 @@
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE DeriveGeneric         #-}
+module Personio.EmployeeStatus (
```
phadej: noticed just now, should it be `Personio.Types.EmployeeStatus`?

personio-client/tests/Tests.hs
```diff
@@ -10,6 +10,8 @@ import Test.Tasty
 import Test.Tasty.HUnit
 import Test.Tasty.QuickCheck
 
+import Personio.EmployeeStatus (Status(..))
```
phadej: Should we re-export that module from `Personio.Types` (which is re-exported from `Personio`), so we don't need this import.

PR #370, opened 2017-07-12
assigned error printing to logger

ArktinenSieni:


proxy-app/src/Futurice/App/Proxy.hs
```diff
@@ -203,8 +203,8 @@ checkCreds ctx req u p = withResource (ctxPostgresPool ctx) $ \conn -> do
         "select 1 from proxyapp.credentials where username = ? and passtext = ?;"
         (u', p') :: IO [Postgres.Only Int]
     case res of
-        [] -> do
-            hPutStrLn stderr $ "Invalid login with: " ++ show (u, p)
+        [] -> runLogT "checkCreds" (ctxLogger ctx) $ do
+            logAttention_ $ "Invalid login with: " <> u' <> ", " <> p'
```
phadej: don't log password.

phadej: Also it's possible to use:
```haskell
logAttention "Invalid login" u' -- underscoreless variant take an `ToJSON`able argument which is logged as well.
```

proxy-app/src/Futurice/App/Proxy.hs
```diff
@@ -193,6 +192,7 @@ defaultMain = futuriceServerMain makeCtx $ emptyServerConfig
             , ctxFumBaseurl           = cfgFumBaseurl
             , ctxFumAuthToken         = cfgFumAuthToken
             , ctxPowerBaseurl         = cfgPowerBaseurl
+            , ctxLogger               = _logger
```
phadej: rename variable to `logger`. Underscore was to prevent "unused variable" warning. Now it's used. I wish there were an automataic way to check "used variables, which look like unused"

phadej:
FWIW, In future, please squash the fixup commits. Now the history is ugly with `moar pretty` commit in between.

PR #371, opened 2017-07-13
Proxy app encpass

ArktinenSieni:
Simplest option as long credentials are added via database.

Instead of

username | passtext
--- | ---
masa | mainio

Will be for example

username | passtext
--- | ---
masa | $1$ayXG/ESL$aT/o0VV5fIi8Ow7G8sxo64

- Will require rebase after ACL has been added.

### DIFF
--- a/proxy-mgmt-app/schema.psql
+++ b/proxy-mgmt-app/schema.psql
@@ -14,16 +14,13 @@ CREATE TABLE proxyapp.accesslog (
   endpoint TEXT NOT NULL
 );
 
-CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
+CREATE FUNCTION proxyapp.pass_crypt() RETURNS trigger AS $pass_crypt$
   BEGIN
-    -- If given passtext is non-empty, encrypt it
-    IF NEW.passtext IS NOT NULL AND NEW.passtext NOT LIKE '' THEN
-      NEW.passtext = crypt(NEW.passtext, gen_salt('md5'));
-    END IF;
+    NEW.passtext = crypt(NEW.passtext, gen_salt('bf'));
 
     RETURN NEW;
   END;
  $pass_crypt$ LANGUAGE plpgsql;
 
 CREATE TRIGGER pass_crypt BEFORE INSERT OR UPDATE ON proxyapp.credentials
-  FOR EACH ROW EXECUTE PROCEDURE pass_crypt();
+  FOR EACH ROW EXECUTE PROCEDURE proxyapp.pass_crypt();
--- a/proxy-mgmt-app/schema.psql
+++ b/proxy-mgmt-app/schema.psql
@@ -12,3 +12,18 @@ CREATE TABLE proxyapp.accesslog (
   updated TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
   endpoint TEXT NOT NULL
 );
+
+CREATE EXTENSION pgcrypto;
+
+CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
+  BEGIN-- If given passtext is non-empty, encrypt it
+    IF NEW.passtext IS NOT NULL AND NEW.passtext NOT LIKE '' THEN
+      NEW.passtext = crypt(NEW.passtext, gen_salt('md5'));
+    END IF;
+
+    RETURN NEW;
+  END;
+ $pass_crypt$ LANGUAGE plpgsql;
+
+CREATE TRIGGER pass_crypt BEFORE INSERT OR UPDATE ON proxyapp.credentials
+  FOR EACH ROW EXECUTE PROCEDURE pass_crypt();
--- a/proxy-mgmt-app/schema.psql
+++ b/proxy-mgmt-app/schema.psql
@@ -1,4 +1,5 @@
 -- CREATE SCHEMA proxyapp;
+CREATE EXTENSION pgcrypto;
 
 CREATE TABLE proxyapp.credentials (
   username TEXT PRIMARY KEY,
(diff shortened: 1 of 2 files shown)


proxy-mgmt-app/schema.psql
```diff
@@ -12,3 +13,17 @@ CREATE TABLE proxyapp.accesslog (
   updated TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
   endpoint TEXT NOT NULL
 );
+
+CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
+  BEGIN
+    -- If given passtext is non-empty, encrypt it
+    IF NEW.passtext IS NOT NULL AND NEW.passtext NOT LIKE '' THEN
+      NEW.passtext = crypt(NEW.passtext, gen_salt('md5'));
```
phadej: does postgres support sha512 scheme?

FYI http://hackage.haskell.org/package/cryptohash-sha512

proxy-mgmt-app/schema.psql
```diff
@@ -12,3 +13,17 @@ CREATE TABLE proxyapp.accesslog (
   updated TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
   endpoint TEXT NOT NULL
 );
+
+CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
+  BEGIN
+    -- If given passtext is non-empty, encrypt it
+    IF NEW.passtext IS NOT NULL AND NEW.passtext NOT LIKE '' THEN
```
phadej: why new passtext can be empty, if it's not modified?

proxy-mgmt-app/schema.psql
```diff
@@ -12,3 +13,17 @@ CREATE TABLE proxyapp.accesslog (
   updated TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
   endpoint TEXT NOT NULL
 );
+
+CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
```
phadej: are functions global, or can they be in schema (`proxyapp`?)

proxy-mgmt-app/schema.psql
phadej: does this work like

```psql
INSERT iNTO credentials (username, password) VALUES ('foo', 'secret');
```

and the secret will be automatically `crypt`ed?

proxy-mgmt-app/schema.psql
ArktinenSieni: It does. I'll change it.

proxy-mgmt-app/schema.psql
ArktinenSieni: Passtext can be empty because I was not sure about following:
`passtext TEXT, -- if null it's disabled`
which was in schema file.

The "secret" would be automatically encrypted. Same applies to UPDATE.

proxy-mgmt-app/schema.psql
ArktinenSieni: I have to check. I have a hunch that the function is global to the database it is created at.
At least in this case the function is applied to the database in the schema `proxyapp`

ArktinenSieni: > If a schema name is included, then the function is created in the specified schema. Otherwise it is created in the current schema.
[Postgres documentation](https://www.postgresql.org/docs/9.1/static/sql-createfunction.html)

Will have to add the schema name.

proxy-mgmt-app/schema.psql
phadej: hmm. Not sure if disabling makes sense. Let's make the field `NOT NULL` and disable via `DELETE` :)

proxy-app/proxy-app.cabal
```diff
@@ -37,11 +37,13 @@ library
     , blaze-html
     , bytestring
     , cassava
+    , cryptohash-sha512
     , env-config
     , fum-client
     , futurice-github
     , futurice-prelude
     , futurice-servant
+    , hex
```
phadej: use `base16-bytestring`

ArktinenSieni:
- encryption from md5 to sha512
- column `passtext` must be now defined
- encryption function is now tied to `proxyapp`-schema

proxy-mgmt-app/schema.psql
```diff
@@ -16,11 +16,7 @@ CREATE TABLE proxyapp.accesslog (
 
 CREATE FUNCTION pass_crypt() RETURNS trigger AS $pass_crypt$
   BEGIN
-    -- If given passtext is non-empty, encrypt it
-    IF NEW.passtext IS NOT NULL AND NEW.passtext NOT LIKE '' THEN
-      NEW.passtext = crypt(NEW.passtext, gen_salt('md5'));
-    END IF;
-
+    NEW.passtext = ENCODE(DIGEST(NEW.passtext, 'sha512'), 'hex');
```
phadej: ah, bummer. I meant:

screenshot from 2017-07-13 17-07-31

`crypt` on linux supports `sha256` based scheme. 

but postgres seems not: https://www.postgresql.org/docs/9.6/static/pgcrypto.html :( bummer.

Sorry, then I guess `crypt(?, 'md5')` is still better than saltless (or even saltful) `sha512`

phadej: Correction, let's go with `bf`. It's more secure than `md5`.

phadej: For future reference: https://en.wikipedia.org/wiki/Crypt_(C)#Blowfish-based_scheme

stack-ghc-8.0.yaml
```diff
@@ -1,16 +1,13 @@
-resolver: nightly-2017-06-01
```
phadej: aargh.

PR #375, opened 2017-07-19
Added employment type, validated external contract end date

ArktinenSieni:
`EmploymentType` was made in similar fashion as `Status`. 
Contract end date should now be present if `EmploymentType` is `External`

### DIFF
--- a/personio-client/src/Personio/Types.hs
+++ b/personio-client/src/Personio/Types.hs
@@ -28,7 +28,9 @@ import Futurice.Prelude
 import Prelude ()
 import Text.Regex.Applicative.Text (RE', anySym, match, psym, string)
 
-import Personio.Types.EmployeeStatus (Status (..))
+import Personio.Types.EmployeeEmploymentType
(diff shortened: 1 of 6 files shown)


personio-client/src/Personio/Types.hs
```diff
@@ -529,6 +538,23 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+                Just External -> checkEndDate
+                _             -> pure ()
+          where
+            checkEndDate = do
+              eDate <- lift (parseAttribute obj "contract_end_date")
+              case eDate of
+                  Null     -> tell [ExternalEndDateMissing]
+                  String d -> checkAttributeName d ExternalEndDateMissing
+                  _        -> pure ()
+
```
phadej: plenty of empty lines here. One would be enough :)

ArktinenSieni: Wops. They're gone now ^^

PR #378, opened 2017-07-19
Absence created

phadej:
Let's merge this later, as it will need redeployment of planmill-proxy and reports-app (at least?)


phadej:
tests failing as expected (data structure hash changes!).

I already forgot I wrote that test :)

PR #384, opened 2017-07-24
encodeUtf8 now imported from Prelude

ArktinenSieni:
Not sure if was required, but also replaced whole `Text.Encode` imports with a individual function import, if only the one function was needed from `Text.Encode`.

### DIFF
--- a/reports-app/src/Futurice/App/Reports.hs
+++ b/reports-app/src/Futurice/App/Reports.hs
@@ -26,10 +26,10 @@ import Numeric.Interval.NonEmpty ((...))
 import Prelude ()
 import Servant
 import Servant.Chart             (Chart (..))
+import Data.Text.Encode          (decodeUtf8)
 
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Text            as T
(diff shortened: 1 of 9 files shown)


reports-app/src/Futurice/App/Reports.hs
```diff
@@ -287,7 +287,7 @@ ctxToIntegrationsConfig now (_cache, mgr, lgr, Config {..}) = MkIntegrationsConf
 repos :: Manager -> Text -> IO [GitHubRepo]
 repos mgr url = do
     req <- parseUrlThrow $ T.unpack url
-    res <- TE.decodeUtf8 . LBS.toStrict . responseBody <$> httpLbs req mgr
+    res <- decodeUtf8 . LBS.toStrict . responseBody <$> httpLbs req mgr
```
phadej: `decodeUtf8Lenient`

phadej:
Also check `stack build --pedantic` (travis helps!)

phadej:
travis still red

phadej:
try `stack build --pedantic` locally (after `stack clean`)

PR #391, opened 2017-07-28
Improve docs

kimmoahokas:
Add commands for installing prerequisites on mac and for creating docker images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@
     - If you didn't clone recursively, fetch the submodules with `git submodule update --init`
 2. Install native dependencies:
     - macOS:
-        - `brew install haskell-stack fftw pkg-config`
+        - Install [haskell stack](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
+        - `brew install fftw pkg-config`
         - `stack setup`
         - `stack install alex happy`
         - https://postgresapp.com/ and `source env-postgres-osx.sh`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,14 +10,21 @@
     - If you didn't clone recursively, fetch the submodules with `git submodule update --init`
 2. Install native dependencies:
     - macOS:
-        - `brew install fftw`
+        - `brew install haskell-stack fftw pkg-config`
```
phadej: no. Don't install stack with `brew`. It will only cause problems.

README.md
```diff
@@ -10,14 +10,21 @@
...
         - https://postgresapp.com/ and `source env-postgres-osx.sh`
     - Ubuntu: `apt-get install libfftw3-dev libpq-dev`
 3. Build and run *theme-app*
     - theme-app is a small app without external integrations.
     - `stack build --fast theme-app`
     - `LOGENTRIES_TOKEN=00000000-0000-0000-0000-000000000000 stack exec -- theme-app-server`
 
+## Building docker images
+
+`stack exec -- haskell-mega-repo-tool build-docker <app-name>`, for example
```
phadej: just `mega-repo-tool`

phadej: `appname` is actually swarm's app name, i.e. the thing in `<appname>.app.futurice.com`, so it's `checklist` without `2`

README.md
kimmoahokas: So it's better to follow instructions at https://docs.haskellstack.org/en/stable/install_and_upgrade/ ? Can you point me to any infromation why installing with homebrew is not good?

phadej: > The Homebrew formula and bottles are **unofficial** and lag slightly behind new Stack releases, but tend to be updated within a day or two.

If something doesn't work, unofficiality causes extra work when reporting issues, `stack` isn't as stable as e.g. `fftw`.

PR #401, opened 2017-08-07
Contacts app personio

phadej:


personio-client/fixtures/employee.json
```diff
@@ -189,6 +189,10 @@
       "label": "Github",
       "value": "https://github.com/gitMastur"
     },
+    "dynamic_72914": {
```
phadej: ping @ArktinenSieni

PR #406, opened 2017-08-08
Personio client prepare

ArktinenSieni:


personio-client/src/Personio/Types.hs
```diff
@@ -136,7 +136,7 @@ data Employee = Employee
     , _employeeFlowdock       :: !(Maybe FD.UserId)
     , _employeeStatus         :: !Status
     , _employeeHRNumber       :: !(Maybe Int)
-    , _employeeEmploymentType :: !EmploymentType
+    , _employeeEmploymentType :: !(Maybe EmploymentType)
```
phadej: do we have a check if employment type is missing?

ArktinenSieni: At least there is a validation for it at row 516.

phadej: :+1:

phadej:
Merge when green.

phadej:
red

PR #418, opened 2017-08-15
Personio client validations

ArktinenSieni:
Most of the validations assume that labels of dynamic values correspond the fields seen in a Personio profile. Would be great to check.

Fields not yet validated:

- Monthly fixed salary 100%
  - requires better specifications to validate
- Monthly variable salary 100%
  - will do tomorrow
- Hourly salary x 100
  - requires better specifications to validate

Following looks like a lot of hassle. Do we want to validate these?
- (SE) Occupational pension %
- (SE) Holidays
- BIC
- (FI) Social Security Number
- (DE) ID number
- (DE) Social security number (SV)
- (GB) National Insurance Number
- (SE) Personal number
- Home postal code
  - Some countries do not even use, and most countries have different valid codes
- Home supervisor
  - cannot be validated here?

### DIFF
--- a/personio-client/src/Personio/Types.hs
+++ b/personio-client/src/Personio/Types.hs
@@ -10,13 +10,14 @@ module Personio.Types (
     module Personio.Types, -- TODO: this exports some unnecessary stuff too.
     module Personio.Types.ContractType,
     module Personio.Types.EmploymentType,
+    module Personio.Types.PersonalIdValidations,
     module Personio.Types.Status,
     ) where
 
 -- Uncomment to get attribute hashmap
 -- #define PERSONIO_DEBUG 1
 
-import Control.Monad.Writer        (WriterT, execWriterT)
+import Control.Monad.Writer        (WriterT, execWriterT, unless)
 import Data.Aeson.Compat
 import Data.Aeson.Internal         (JSONPathElement (Key), (<?>))
 import Data.Aeson.Types
@@ -40,6 +41,7 @@ import Text.Regex.Applicative.Text (RE', anySym, match, psym, string)
 
 import Personio.Types.ContractType
 import Personio.Types.EmploymentType
+import Personio.Types.PersonalIdValidations
 import Personio.Types.Status
 
 import qualified Chat.Flowdock.REST            as FD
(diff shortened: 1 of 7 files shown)


phadej:
FWIW.
-  `FI Social Security Number ` validation isn't any complicated then IBAN number, so we can have those. I guess there's some validations for other numbers as well.
- BIC isn't that important, I'm not sure if there's "free" database for them
- postal code and supervisor aren't worth validating atm.

personio-client/src/Personio/Types.hs
```diff
@@ -641,6 +697,86 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
+        hoursValidate = do
+            hours <- lift (parseAttribute obj "weekly_working_hours")
+            case toNum hours of
+                Just n  -> if n > 0
+                    then pure ()
+                    else tell [HoursInvalid n]
+                Nothing -> lift (typeMismatch "weekly_working_hours" hours)
+          where
+            toNum :: Value -> Maybe Scientific
+            toNum (String v) = readMaybe $ T.unpack v
```
phadej: is it really both string and num in real data?

ArktinenSieni: Was being pessimistic since in current fixtures weelky_working_hours is set as string ("37.5")

ArktinenSieni: It is written as String which makes no sense to me. Should I leave "Number"-possibility in case it will be changed? @phadej

phadej: I hope it won't ever, so let it just fail in that case.

ArktinenSieni:
Now starting with social security numbers etc.

personio-client/src/Personio/Types.hs
```diff
@@ -742,7 +746,34 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
                     dynamicAttributeMissing "Nationality" NationalityMissing
                     dynamicAttributeMissing "Work permit" WorkPermitMissing
                     dynamicAttributeMissing "Career path level" CareerPathLevelMissing
+                    sweValidations
                 _             -> pure ()
+          where
+            sweValidations = do
+                nat <- lift (parseDynamicAttribute obj "Nationality")
+                case nat of
+                    String n -> if T.toLower n /= "sweden"
```
phadej: `when`

ArktinenSieni:
My validation on Swedish personal id number might be incorrect, since using format YYMMDDCCCS
 and YYYYMMDDCCCS caused as many validation errors on real data. I used this in validation: https://en.wikipedia.org/wiki/Personal_identity_number_(Sweden)
Am I even using the right validation? ^^'

personio-client/src/Personio/Types.hs
```diff
@@ -863,7 +867,21 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
...
-                _ -> lift (typeMismatch "Nationality" nat)
+                _        -> lift (typeMismatch "Nationality" nat)
+
+        sePINValidate :: WriterT [ValidationMessage] Parser ()
+        sePINValidate = do
+            nat <- lift (parseDynamicAttribute obj "Nationality")
+            pin <- lift (parseDynamicAttribute obj "(SE) Personal number")
+            case (nat, pin) of
+                (String n, Number p) -> if T.toLower n == "sweden"
+                    then when (not $ validPIN p) $ tell [SEPersonalIdInvalid]
```
phadej: `when (not ...)` = `unless ...`

personio-client/src/Personio/Types.hs
```diff
@@ -863,7 +867,21 @@ validatePersonioEmployee = withObjectDump "Personio.Employee" $ \obj -> do
                 String n -> if T.toLower n == "finland"
                     then when (not $ isValidFinSSN ssn) $ tell [FISSNInvalid]
                     else pure ()
-                _ -> lift (typeMismatch "Nationality" nat)
+                _        -> lift (typeMismatch "Nationality" nat)
+
+        sePINValidate :: WriterT [ValidationMessage] Parser ()
+        sePINValidate = do
+            nat <- lift (parseDynamicAttribute obj "Nationality")
+            pin <- lift (parseDynamicAttribute obj "(SE) Personal number")
```
phadej: is it really stored as number? (Hopefully it cannot have significant leading zeros!)

personio-client/src/Personio/Types/PersonalIdValidations.hs
```diff
@@ -0,0 +1,137 @@
...
+
+-- | Validate Swedish Personal Identity Number
+-- 
+-- See <https://en.wikipedia.org/wiki/Personal_identity_number_(Sweden)>
+isValidSwePIN :: Integer -> Bool
+isValidSwePIN = isValidPINString . fromIntegral
+  where
+    isValidPINString :: Integer -> Bool
+    isValidPINString p =
+        let pin    = numToDigits p
```
phadej: I'd use regex with `range` you already have somewhere

personio-client/src/Personio/Types.hs
ArktinenSieni: Was incorrectly set as number, which caused lots of confusion. Will be corrected on the other end to String instead.

personio-client/tests/Tests.hs
```diff
@@ -314,6 +314,12 @@ validations = testGroup "Validations"
        $ correctEmployeeValue
            & attributeValue "dynamic_66601" . _String .~ "Finland" -- Nationality
```
phadej: How non-Finns who have *hetu* are represented. Won't residence be more correct than nationality. How it works in Sweden&Germany?

ArktinenSieni: Followed specifications set in the implementation plan. I can be more specific with validations if needed. 

- Finnish SSN is issued to people who stay extended time periods in Finland, and automatically if born in Finland
- Swedish personal number is set from birth, and only changed in case of errors. Persons with no Swedish SSN can receive a separate number
- German SSN is applied to, and is set if one is part of German social security system.
- German ID number, which seems to be the tax number, can be sought and is needed if one gets salary in Germany.

ArktinenSieni: Will ask from HC-team if more validations to these fields are wanted

ArktinenSieni:
On 25.8, all fields but "Home country" are validated following instructions from Personio [implementation plan](https://docs.google.com/spreadsheets/d/1RxdJze-NT4DLZBxIQqgPwqnHot5YeVP7P6wMQ1lXbAE/edit#gid=1970701317).

ArktinenSieni:
Field "Home country" is required to be set. 
Validation of home country (by instructions set in the implementation plan) is not yet implemented.

PR #477, opened 2017-11-08
Servant-0.12 changes

phadej:


phadej:
This for sure breaks `stack.yaml`s. WIll fix when someone notices :)

PR #494, opened 2017-12-13
Update docs

phadej:


phadej:
@kimmoahokas could you check `README.md` changes. That's what we did today, did I miss something?

phadej:
If build is `ByteString.Builder`, then it would be [`fromEncoding`](http://hackage.haskell.org/package/aeson-1.2.3.0/docs/Data-Aeson-Encoding.html#v:fromEncoding)`. toEncoding`.

Naming is a little confusing, blame me for that.

If, OTOH, you wont to get `text`'s [`Builder`](http://hackage.haskell.org/package/text-1.2.2.2/docs/Data-Text-Lazy-Builder.html#t:Builder), then AFAIK `aeson` doesn't have it.

PR #497, opened 2018-01-11
Add ....

neurogoo:
Add new color transformer


phadej:
Ah, you should commit `futurice-logo/futurice-logo.cabal` too.

phadej:
There's some merge commit. Could you redo the branch without it.

We try to avoid merge commits in so called feature branches, and use rebase if we need to get new stuff from `master`. You can force push into branches you work on.

PR #499, opened 2018-01-16
Base for having user preferences in hours app

neurogoo:
Give user a way to load and save user preferences. Current setting list is not final


hours-api/src/Futurice/App/HoursApi.hs
```diff
@@ -67,6 +71,25 @@ authorisedUser ctx mfum meterName action =
...
+  :: Ctx
+  -> Maybe FUM.Login
+  -> Handler [SettingsResponse]
+settingsHandler ctx mfum =
+  mcase (mfum <|> ctxMockUser ctx) (throwError err403) $ \fumUsername -> do
+    liftIO $ mark "Request settings"
+    liftIO $ runLogT "logic" (ctxLogger ctx) $ settingsEndpoint ctx fumUsername
+
+updateSettingsHandler
+  :: Ctx
```
phadej: 4 spaces.

Somehow I manage to start with most important things first :S

hours-api/src/Futurice/App/HoursApi.hs
```diff
@@ -67,6 +71,25 @@ authorisedUser ctx mfum meterName action =
...
+  -> Maybe FUM.Login
+  -> SettingsResponse
+  -> Handler SettingsUpdateResponse
+updateSettingsHandler ctx mfum set =
+    mcase (mfum <|> ctxMockUser ctx) (throwError err403) $ \fumUsername -> do
```
phadej: 3 -> 4

hours-api/src/Futurice/App/HoursApi.hs
```diff
@@ -67,6 +71,25 @@ authorisedUser ctx mfum meterName action =
...
+     liftIO $ mark "Request edit settings"
```
phadej: 4 -> 8.

FWIW, have you glanced through linked https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md in https://github.com/futurice/haskell-mega-repo#style-guide

We probably deviate a little, and I have to update our guide to pinpoint these differences.
(i.e. we indent top-level as you did, except with 4 spaces, and not as in Johan's example).

hours-api/src/Futurice/App/HoursApi.hs
```diff
@@ -67,6 +71,25 @@ authorisedUser ctx mfum meterName action =
...
+     liftIO $ runLogT "logic" (ctxLogger ctx) $ updateSettingsEndpoint ctx fumUsername set
```
phadej: `set` is bad name. I'd use `sr` (for `SettingsResponse`).

hours-api/src/Futurice/App/HoursApi.hs
phadej: as Everything happens in IO, i'd change

```haskell
foo = do
    liftIO bar
    liftIO baz
```

to

```haskell
foo = liftIO $ do
    bar
    baz
```

hours-api/src/Futurice/App/HoursApi.hs
```diff
@@ -97,6 +120,9 @@ defaultMain = futuriceServerMain (const makeCtx) $ emptyServerConfig
         let pmCfg = cfgPlanmillCfg config
         ws <- PM.workers lgr mgr pmCfg ["worker1", "worker2", "worker3"]
 
+        let psqlCfg = cfgPostgresConnInfo config
+        pp <- createPostgresPool psqlCfg
```
phadej:
```haskelL
pp <- createPostgresPool (cfgPostgresConnInfo config)
```

FWIW, I'd should inlined `pmCfg` above too. or used also in `Ctx` construction below. My bad.

hours-api/src/Futurice/App/HoursApi/Config.hs
```diff
@@ -9,17 +9,20 @@ import Futurice.Integrations
 import Futurice.Prelude
 import Prelude ()
 
+import           Database.PostgreSQL.Simple (ConnectInfo)
 import qualified FUM
-import qualified PlanMill as PM
+import qualified PlanMill                   as PM
```
phadej: we keep `import` and `import qualified` sections separated by line, to prevent this kind of not-nice formatting.

hours-api/src/Futurice/App/HoursApi/Ctx.hs
```diff
@@ -20,4 +21,8 @@ data Ctx = Ctx
     , ctxWorkers         :: !PM.Workers
     , ctxPlanmillCfg     :: !PM.Cfg
     , ctxIntegrationsCfg :: !(IntegrationsConfig '[I, I, Proxy, Proxy, Proxy, Proxy])
+    , ctxPostgresPool    :: !(Pool Connection)
     }
+
+instance HasPostgresPool Ctx where
+  postgresPool = ctxPostgresPool
```
phadej: 4 spaces

hours-api/src/Futurice/App/HoursApi/Logic.hs
```diff
@@ -12,20 +12,26 @@ module Futurice.App.HoursApi.Logic (
     entryEndpoint,
     entryEditEndpoint,
     entryDeleteEndpoint,
+    settingsEndpoint,
+    updateSettingsEndpoint,
     ) where
 
 import Control.Lens              (maximumOf, (<&>))
+import FUM.Types.Login           (Login)
+import Futurice.App.HoursApi.Ctx
```
phadej: We should reading/writing settings functions into `MonadHours` type-class. So the `HoursApi.Ctx` implementation detail don't leak into `Logic`, but stays in `Monad`.

So far settings are only view related, so don't affect other endpoints, but they could in future.

phadej:
Quite many style related comments, but also one architectural one.

neurogoo:
Next MonadHours refractoring

neurogoo:
A little unsure if editSettings is done right

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -296,6 +348,10 @@ instance MonadHours Hours where
         reports <- cachedPlanmillAction (PM.timereportsFromIntervalFor resultInterval pmUid)
         traverse convertTimereport (toList reports)
 
+    settings = Hours $ lift $ Haxl.dataFetch FetchSettings
+
+    editSettings a b = void $ Hours $ lift $ Haxl.dataFetch $ EditSettings a b
```
phadej: https://hackage.haskell.org/package/haxl-0.5.1.0/docs/Haxl-Core.html#v:uncachedRequest

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -154,7 +160,53 @@ instance NFData a => NFData (Wrapped a) where
...
+                    "SELECT weekly_view, show_graphs FROM hours.preferences WHERE username IN ?;" $ Postgres.Only login :: IO [T.SettingsResponse]
+            case listToMaybe res of
+              (Just s) -> case s == old of
+                False -> Haxl.putFailure v $ PostgresException "Data changed in database"
+                True -> do
+                    _ <- runLogT "editSettings" logger $ safePoolExecute pool "UPDATE hours.preferences SET weekly_view=?, show_graphs=? WHERE username = ?;" $
+                        searchParams new login
+                    Haxl.putSuccess v new
+                      where searchParams (T.SettingsResponse w s) login = (w,s,login)
+              Nothing -> undefined
```
phadej: undefined...

I'd rather issue an `UPDATE` directly, requiring `weekly_view`, and `show_graphs` to match old settings. Than based on how many rows are updated, deduce whether the update was successful.

Maybe we need upsert here to deal with *no settings* case.

hours-api/src/Futurice/App/HoursMock/Monad.hs
```diff
@@ -144,3 +145,9 @@ instance MonadHours Hours where
     editTimereport tid ntr = do
         deleteTimereport tid
         addTimereport ntr
+
+    settings = return SettingsResponse
+                      { _settingsResponseWeeklyView = False
```
phadej: indentation is too far to the right :)

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -154,7 +160,53 @@ instance NFData a => NFData (Wrapped a) where
...
+    FetchSettings :: PostgresRequest T.SettingsResponse
+    EditSettings :: T.SettingsResponse -> T.SettingsResponse -> PostgresRequest T.SettingsResponse
+
+deriving instance Show (PostgresRequest a)
+deriving instance Typeable PostgresRequest
+deriving instance Eq (PostgresRequest a)
+
+instance Hashable (PostgresRequest a) where
+    hashWithSalt salt FetchSettings = salt `hashWithSalt` (0::Int)
+    hashWithSalt salt (EditSettings a b) = salt `hashWithSalt` (1::Int,a,b)
```
phadej: hmm, neat. I always wrote those as

```haskell
salt `hashWithSalt` (1 :: Int) `hashWithSalt` a `hashWithSalt` b
```

But this is way more compact! I'd add spaces after the comma though: `(1 :: Int, a, b)`

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -154,7 +160,53 @@ instance NFData a => NFData (Wrapped a) where
...
+
+instance Haxl.StateKey PostgresRequest where
+    data State PostgresRequest = PostgresState Logger Login (Pool Connection)
+
+instance Haxl.DataSourceName PostgresRequest where
+  dataSourceName _ = "PostgresRequest"
```
phadej: 4

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -154,7 +160,53 @@ instance NFData a => NFData (Wrapped a) where
     rnf (Wrap (Right x))                = rnf x
     rnf (Wrap (Left (SomeException e))) = seq e ()
 
-
+data PostgresRequest a where
+    FetchSettings :: PostgresRequest T.SettingsResponse
+    EditSettings :: T.SettingsResponse -> T.SettingsResponse -> PostgresRequest T.SettingsResponse
```
phadej: can this be rather `Bool` to indicate if it succeeded, I don't think we need to return `new`. But we probably want 
recovery to be possible (and not abort with an exception).

phadej:
Now, as we have preference app, it would make sense to store `hours` preferences there. Then we can keep `hours-api` stateless (good!)

neurogoo:
Sounds good! I will look into it

neurogoo:
Actually I don't think that this branch has a much of value now with preferences app doing mostly the same thing. I would just delete this branch and open new one when this feature is actually needed

PR #503, opened 2018-01-29
Add absence reason to absence task names

neurogoo:
Check if task is absence task and if it is change the name to absence reason. Convert should be enumeration call, but couldn't think of clean way to do it

### DIFF
--- a/hours-api/src/Futurice/App/HoursApi/Class.hs
+++ b/hours-api/src/Futurice/App/HoursApi/Class.hs
@@ -44,6 +44,11 @@ module Futurice.App.HoursApi.Class (
     raFinish,
     raProjectId,
     raTaskId,
+    Absence (..),
+    absenceProject,
+    absenceStart,
+    absenceFinish,
+    absenceAbsenceType,
     ) where
 
 import Data.Fixed                (Centi)
@@ -138,6 +143,11 @@ class (MonadTime m) => MonadHours m where
         today <- currentDay
         timereports (addDays (-28) today ... today)
 
+    -- | Absences
+    --
+    -- Return all absences
+    absences :: Interval Day -> m [Absence]
+
 -------------------------------------------------------------------------------
 -- Data
 -------------------------------------------------------------------------------
@@ -145,7 +155,7 @@ class (MonadTime m) => MonadHours m where
 data ReportableAssignment = ReportableAssignment
     { _raProjectId :: !PM.ProjectId
     , _raTaskId    :: !PM.TaskId
-    , _raFinish    :: !UTCTime
+    , _raFinish    :: !Day
     }
   deriving (Eq, Show, Generic)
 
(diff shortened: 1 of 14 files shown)


hours-api/src/Futurice/App/HoursApi/Class.hs
```diff
@@ -154,6 +155,7 @@ data Task = Task
     , _taskName      :: !Text
     , _taskProjectId :: !PM.ProjectId
     , _taskFinish    :: !UTCTime
+    , _taskDutyType  :: !(Maybe Int)
```
phadej: Here we want  `Text`

planmill-client/src/PlanMill/Types/Absence.hs
```diff
@@ -30,6 +30,7 @@ data Absence = Absence
     , absenceProject          :: !ProjectId
     , absenceStart            :: !Day
     , absenceFinish           :: !Day
+    , absenceAbsenceType      :: !Int
```
phadej: and here we want `Enum`

hours-api/src/Futurice/App/HoursApi/Class.hs
```diff
@@ -194,6 +203,15 @@ data Capacity = Capacity
     }
   deriving (Eq, Show, Generic)
 
+data Absence = Absence
+    { _absencePerson           :: !PM.UserId
```
phadej: do we need this? Can we return only "my" absences? i.e. filter out others absences already `Monad`

hours-api/src/Futurice/App/HoursApi/Monad.hs
```diff
@@ -297,6 +296,23 @@ instance MonadHours Hours where
...
+                  text <- absenceTypeToText a
+                  pure Absence
+                      { _absencePerson      = PM.absencePerson a
+                      , _absenceProject     = PM.absenceProject a
+                      , _absenceStart       = PM.absenceStart a
+                      , _absenceFinish      = PM.absenceFinish a
+                      , _absenceAbsenceType = text
+                      }
+              absenceTypeToText :: PM.Absence -> Hours Text
+              absenceTypeToText a = PMQ.enumerationValue (PM.absenceAbsenceType a) (fromString "Unknown Contract")
```
phadej: "Unknown absence type"

planmill-client/src/PlanMill/Types/Enumeration.hs
```diff
@@ -84,14 +83,18 @@ instance FromEnvVarList (EnumValue entity field) where
 
 newtype IM a = IM { getIM :: IntMap a }
 
+
+-- Filtering non-Int:s away as meta json sometimes have useless "optgroup"
```
phadej:
```
-- | 
```

to make that haddock visible comment. (minor thing, mostly FYI)

PR #513, opened 2018-02-15
Small things 2018 02 15

phadej:


neurogoo:
Is there point of having those "hasDifferentPersonio..." if you show the Personio information in separate table

PR #514, opened 2018-02-16
Add absencesFromIntervalFor

phadej:
Requesting absencecs for user drops from 7.667166s to 2.637219s


phadej:
Than it breaks. But they could remove also documented endpoints, so I don't see a risk here :)

neurogoo:
What is preventing using ApplicativeDo?

phadej:
Hour *marking* breaks, I don't remember details right now :/

PR #540, opened 2018-03-19
Remove the need of fetching employee list from FUM

neurogoo:


planmill-sync/src/Futurice/App/PlanMillSync.hs
```diff
@@ -27,10 +25,9 @@ import Futurice.App.PlanMillSync.IndexPage
 import Futurice.App.PlanMillSync.Monad
 import Futurice.App.PlanMillSync.Types
 
-import qualified Data.Map.Strict as Map
-import qualified Data.Set        as Set
+import qualified Data.Set as Set
 import qualified FUM
```
phadej: could we change that to `FUM.Types.Login` (and drop `fum-client` dependency in favor of `fum-types`)

PR #565, opened 2018-04-27
Improve task sorting in checklist page

neurogoo:
Task sorts by:
- First order by dependency
- Then weight each vertex by minimum offset of all of it's "child's"
- Then reorder list based on those weight

A somewhat imperative solution, but seems to work ^^'

### DIFF
--- a/checklist-app/src/Futurice/App/Checklist/Types.hs
+++ b/checklist-app/src/Futurice/App/Checklist/Types.hs
@@ -88,6 +88,7 @@ module Futurice.App.Checklist.Types (
     worldTaskItems',
     worldTasksSorted,
     worldTasksSortedByName,
+    tasksSorted,
     -- * Access
     AuthUser,
     authUserTaskRole,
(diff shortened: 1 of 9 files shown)


checklist-app/src/Futurice/App/Checklist/Types.hs
```diff
@@ -88,6 +88,7 @@ module Futurice.App.Checklist.Types (
     worldTaskItems',
     worldTasksSorted,
     worldTasksSortedByName,
+    worldTasksSortedByOffset,
```
phadej: just change `worldTasksSorted`

checklist-app/src/Futurice/App/Checklist/Types/World.hs
```diff
@@ -108,6 +111,42 @@ worldTasksSorted tr = getter $ \world ->
     sortOn (view taskOffset) $
     Graph.revTopSort (world ^. worldTasks)
 
+type WeightState = State.State (Map (Identifier Task) Integer)
```
phadej: Do you need a monad, can't you build `Map (Identifier Task) Integer` as preprocess step and then `sortOn (\tid -> weights ^? ix tid)` ?

neurogoo: Hmmm, I think that at least I need mapM to guarantee that weights are calculated from left to right as child node min weight effects parent node min weight.

PR #575, opened 2018-05-28
Move schedule-info endpoint from fum-carbon to personio-proxy

neurogoo:
Also added employeeId to payload for later use. Remember to update Schedule url before moving this change to production


phadej:
Looks good. Let's coordinate the deployment on Tuesday

PR #578, opened 2018-06-04
Add new endpoint for deleting multiple timereports at once

neurogoo:
TimereportDelete type needed extra data type to get ToJson and FromJson to work. Why is that and can it be fixed some other way? I think there was similar problem previously, but can't remember where

### DIFF
--- a/hours-api/src/Futurice/App/HoursApi/API.hs
+++ b/hours-api/src/Futurice/App/HoursApi/API.hs
@@ -29,6 +29,7 @@ type FutuhoursV1API =
     :<|> "entry" :> SSOUser :> ReqBody '[JSON] EntryUpdate :> Post '[JSON] EntryUpdateResponse
     :<|> "entry" :> SSOUser :> Capture "id" PM.TimereportId :> ReqBody '[JSON] EntryUpdate :> Put '[JSON] EntryUpdateResponse
     :<|> "entry" :> SSOUser :> Capture "id" PM.TimereportId :> Delete '[JSON] EntryUpdateResponse
+    :<|> "delete-timereports" :> SSOUser :> ReqBody '[JSON] TimereportDelete :> Post '[JSON] EntryUpdateResponse
 
 type FutuhoursAPI = Get '[JSON] Text
     :<|> "api" :> "v1" :> FutuhoursV1API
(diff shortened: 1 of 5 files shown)


hours-api/src/Futurice/App/HoursApi/API.hs
```diff
@@ -29,6 +29,7 @@ type FutuhoursV1API =
     :<|> "entry" :> SSOUser :> ReqBody '[JSON] EntryUpdate :> Post '[JSON] EntryUpdateResponse
     :<|> "entry" :> SSOUser :> Capture "id" PM.TimereportId :> ReqBody '[JSON] EntryUpdate :> Put '[JSON] EntryUpdateResponse
     :<|> "entry" :> SSOUser :> Capture "id" PM.TimereportId :> Delete '[JSON] EntryUpdateResponse
+    :<|> "delete-timereports" :> SSOUser :> ReqBody '[JSON] TimereportDelete :> Post '[JSON] EntryUpdateResponse
```
phadej: Why not simply `[PM.TimereportId]` ? i.e. why wrap in single-field object?

hours-api/src/Futurice/App/HoursApi/Logic.hs
```diff
@@ -94,6 +95,19 @@ entryDeleteEndpoint eid = do
...
+entryDeleteMultipleEndpoint :: H.MonadHours m => TimereportDelete -> m EntryUpdateResponse
+entryDeleteMultipleEndpoint (TimereportDelete eids _) = do
+    let deleteTimereport eid = do
+            tr <- H.timereport eid
+            _ <- H.deleteTimereport eid
+            pure tr
+        days = sort . map (^. H.timereportDay)
+        firstDay = head . days
+        lastDay = last . days
+    trs <- mapM deleteTimereport eids
```
phadej:
```haskell
for_ eids $ \eid -> do
    tr <- H.timereport eid
    _ <- H.depeteTimereport eid
    pure tr
EntryUpdateResponse <$> ...
```

---

What if `eids` is empty? Use `NonEmpty`, not list.

hours-api/src/Futurice/App/HoursApi/API.hs
neurogoo: It felt that if would be a little clearer, but I can change that to that :)

neurogoo:
Feels somewhat clumsy solution. There must be a better way to write it?

PR #583, opened 2018-07-02
Sisosota

phadej:


phadej:
TODO: rewrite with servant-generic

PR #584, opened 2018-07-10
WIP: Futushelf 2.0 aka Library app

neurogoo:
First mvp version. Backend missing snatch(aka forced loan), seeing loans of single person and way to return book. Few comments about code:

Currently most pressing concerns are:
1. How to interface with database. Now done more traditionally (table structure almost straight copy from the old library app). Is this okay or would event sourcing way of working(like in checklist) be better?
2. Is the overall structure of the app anywhere near decent? 
3. Frontend. Needs at least a little bit dynamic magic for main page book pagination. Probably need to use some lines of javascript as anything else (elm, purescript etc.) would probably be overkill

Also currently reads cover pictures from list. This will be changed to using sisosota.

Reason for the abstraction of Loanable type and sum functions are that there is an idea of extending library to track other "loanable" things for example cameras

Also, sorry for the tears you get for reading this code ^^'


library-app/src/Futurice/App/Library/API.hs
```diff
@@ -0,0 +1,38 @@
...
+import Codec.Picture             (DynamicImage)
+import Futurice.Lucid.Foundation (HtmlPage)
+import Futurice.Servant          (SSOUser)
+import Servant.API
+import Servant.HTML.Lucid        (HTML)
+import Servant.JuicyPixels       (PNG)
+
+import Futurice.App.Library.Types
+
+type LibraryAPI =
```
phadej: let's use generics functionality. Check `sisosota` or `avatar` for examples. Or/and: http://haskell-servant.readthedocs.io/en/stable/cookbook/generic/Generic.html

library-app/src/Futurice/App/Library/Types/Library.hs
```diff
@@ -0,0 +1,58 @@
...
+{-# LANGUAGE TypeOperators     #-}
+module Futurice.App.Library.Types.Library where
+
+import Database.PostgreSQL.Simple.FromField as Postgres
+
+import Futurice.Generics
+import Futurice.Prelude
+import Prelude ()
+
+data Library
```
phadej: should it rather be `data Library = OfficeLibrary Office | ELibrary | UnknownLibrary` ?

preferences-app/src/Futurice/App/Preferences/Main.hs
```diff
@@ -22,6 +22,7 @@ import Futurice.Servant
 import Prelude ()
 import Servant
 
+
```
phadej: :)

library-app/src/Futurice/App/Library/API.hs
phadej: with colors: https://haskell-servant.readthedocs.io/en/release-0.14/cookbook/generic/Generic.html

library-app/src/Futurice/App/Library/Types/Library.hs
neurogoo: Ah yes, I didn't remember that we have offices in constants. Good idea

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+
+CREATE TABLE library.library_book (
+    id bigint NOT NULL,
+    information_id bigint NOT NULL,
+    library_id bigint NOT NULL
+);
+
+CREATE TABLE library.library_bookinformation (
+    id bigint NOT NULL,
+    title character varying(255) NOT NULL,
```
phadej: is there a reason we cannot just have `text`?

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
+CREATE SCHEMA library;
+
+CREATE TABLE library.library_book (
+    id bigint NOT NULL,
```
phadej: optimistic type ;)

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    information_id bigint NOT NULL,
+    library_id bigint NOT NULL
```
phadej: I'd use `text` directly. No need to introduce non-natural integral key. Library doesn't have performance concerns.

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
+CREATE SCHEMA library;
+
+CREATE TABLE library.library_book (
```
phadej: Is the idea is that we have `many-to-many` relation between libraries and book? Maybe add a comment to the table.

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    library_id bigint NOT NULL
+);
+
+CREATE TABLE library.library_bookinformation (
+    id bigint NOT NULL,
+    title character varying(255) NOT NULL,
+    isbn character varying(255) NOT NULL,
+    author character varying(255) NOT NULL,
+    publisher character varying(255) NOT NULL,
+    published bigint NOT NULL,
```
phadej: if it's date, use `date`

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
+CREATE SCHEMA library;
+
+CREATE TABLE library.library_book (
+    id bigint NOT NULL,
+    information_id bigint NOT NULL,
+    library_id bigint NOT NULL
+);
+
+CREATE TABLE library.library_bookinformation (
+    id bigint NOT NULL,
```
phadej: primary key?

library-app/schema.sql
phadej: Also you should add foreign keys.

library-app/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    amazon_link character varying(1024)
+);
+
+CREATE TABLE library.library_library (
+    id bigint NOT NULL,
+    name character varying(255) NOT NULL
+);
+
+CREATE TABLE library.library_loan (
+    id bigint NOT NULL,
```
phadej: I'd prefer having `loan_id` etc. names for identifiers, as then you won't mess them that easily.

library-app/schema.sql
neurogoo: Not sure what to you mean with `many-to-many` in this context, but the idea is the that there can be multiple copies of the same book (so same information_id for multiple books) and in same library there can be multiple copies of the same book(so same library_id for multiple books)

library-app/schema.sql
neurogoo: Do you mean in information or in library? For bookinformation I think we should still have separate table as if you change information for book, you want to be sure that it is changed on all the copies of the book. For library, I agree though would some enum be better than text?

library-app/schema.sql
neurogoo: published should be just year and I don't think there are date type that has just year so probably `smallint` is the best option?

phadej: `publishedYear`

library-app/schema.sql
```diff
@@ -0,0 +1,55 @@
+CREATE SCHEMA library;
+
+CREATE TABLE library.library (
```
phadej: As I said previously, let's just have "library_name" as a field in other tables.

It will remove a need to insert a row when there's new office, one less place to update.

library-app/schema.sql
```diff
@@ -0,0 +1,55 @@
...
+);
+
+CREATE TABLE library.item (
+    item_id serial primary key,
+    item_type text NOT NULL,
+    library_id integer references library.library(library_id)
+);
+
+CREATE TABLE library.old_loan (
+    oldloan_id serial primary key,
```
phadej: make a sequence, so `old_loan` and `loan` will share the id.

library-app/schema.sql
```diff
@@ -0,0 +1,55 @@
...
+    loan_id serial primary key,
+    date_loaned date NOT NULL,
+    personio_id integer NOT NULL,
+    item_id serial references library.item(item_id)
+);
+
+CREATE TABLE library.bookinformation (
+    bookinfo_id serial primary key,
+    title text NOT NULL,
+    isbn text NOT NULL,
```
phadej: can `isbn` be the primary key, if it's required (= `NOT NULL`)?

library-app/schema.sql
```diff
@@ -0,0 +1,55 @@
...
+    author text NOT NULL,
+    publisher text NOT NULL,
+    publishedYear smallint NOT NULL,
+    cover text,
+    amazon_link text
+);
+
+CREATE TABLE library.otherinformation (
```
phadej: I don't get the need for this table. Is this games etc.? Please comment.

library-app/schema.sql
```diff
@@ -0,0 +1,55 @@
...
+    otherinfo_id serial primary key,
+    name text NOT NULL
+);
+
+CREATE TABLE library.item_to_bookinformation (
+    item_id serial references library.item(item_id),
+    bookinfo_id serial references library.bookinformation(bookinfo_id),
```
phadej: So `item` might be an instantiation of `bookinformation` ?

Can't we rather have a foreign key (not `NOT NULL`) to `bookinformation`.

And a `CHECK` that exactly one of `bookinfo_id` or `otherinfo_id` is not null.

library-app/schema.sql
neurogoo: This was mostly for "Other" category, things that don't fit to any other category. But I can remove it and add for example games category.

library-app/schema.sql
neurogoo: Well we lose the guarantees that library names are written the same for all, but I can remove it

library-app/schema.sql
neurogoo: Sure why not

library-app/schema.sql
neurogoo: Hmm I am not sure what we get by trying to make the foreign key constraint not null? As every book should have bookinformation. Unique checking across multiple tables seems to need a little bit of SQL black magic. But I will see if I can do it

library-app/schema.sql
```diff
@@ -37,19 +14,35 @@ CREATE TABLE library.bookinformation (
...
+    personio_id integer NOT NULL,
+    item_id integer NOT NULL references library.item(item_id)
 );
 
-CREATE TABLE library.item_to_otherinformation (
-    item_id serial references library.item(item_id),
-    otherinfo_id serial references library.otherinformation(otherinfo_id),
-    UNIQUE(item_id, otherinfo_id)
+CREATE TABLE library.old_loan (
+    oldloan_id integer primary key default nextval('library.loan_seq'),
```
phadej: I think we should have default here. As `loan`s are moved to `old_loan`; default would allow us to accidentally forget

neurogoo: You probably mean should not have? Sure, that sounds reasonable.

library-app/schema.sql
```diff
@@ -1,5 +1,6 @@
 CREATE SCHEMA library;
 
+
```
phadej: ?

preferences-app/src/Futurice/App/Preferences/Main.hs
phadej: please avoid these

library-app/schema.sql
neurogoo: Heh, merged with Github's tool and that didn't show that there was extra space.

phadej:
Merged manually

PR #588, opened 2018-07-16
Library schema

phadej:
I wrote some pseudo code, and noticed few things (see my commit).


phadej:
Schema is almost the same, the changes are visible in my commits.
- UNIQUE `item_id` in `loan` table
- `old_loan.loan_id` has no default
- `item` doesn't have `item_type`, it's obvious from foreign keys

PR #592, opened 2018-07-18
Update datatypes and functions operating them to current database schema

neurogoo:


library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -19,152 +18,150 @@ import qualified Data.Map  as Map
...
+    { idItemId                 :: !ItemId
+    , idLibrary                :: !Library
+    , idBookInformationId      :: !(Maybe BookInformationId)
+    , idBoardGameInformationId :: !(Maybe BoardGameInformationId)
+    } deriving (Generic, FromRow)
+
+data LoanData = LoanData
+    { ldLoanId        :: !LoanId
+    , ldDateLoaned    :: !Day
+    , ldPersonioId    :: !Int32
```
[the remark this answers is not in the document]
neurogoo: There wasn't instance for FromField for it so I thought it wasn't worth the effort. But I will write instance for it

library-app/src/Futurice/App/Library/Types/Item.hs
```diff
@@ -0,0 +1,41 @@
...
+data Item = Item
+    { _itemId      :: !ItemId
+    , _itemLibrary :: !Library
+    , _itemInfo    :: !ItemInfo
+    }
+    deriving (Show, Generic)
+
+data ItemInfo = ItemBook BookInformation
+              | ItemBoardGame BoardGameInformation
+              | ItemUnknown Text
```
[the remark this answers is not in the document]
neurogoo: As ItemData has Maybe fields for all info fields, I think you need either this or just throw error for totalitaly. Would throwing error be better? And I guess there is always small chance that you don't find information for some item_id if as that is enforced just with Check

library-app/src/Futurice/App/Library/Logic.hs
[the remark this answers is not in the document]
neurogoo: Is there actually good way to convert from ByteString to Word?

library-app/src/Futurice/App/Library/Types/Item.hs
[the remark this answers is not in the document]
phadej: We can and should rely on the `CHECK`.

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -148,20 +191,18 @@ bookIdToInformation iid bookidmap bookinfomap = do
...
-           Just (_, lib) -> lib
-           Nothing -> UnknownLibrary
-   bookIdToItem bs bmap iid = case bookIdToInformation iid bs bmap of
-       Just info -> Item iid (bookLibrary iid bs) (ItemBook info)
-       Nothing -> Item iid UnknownLibrary (ItemUnknown "Couldn't find information for book")
-   loans' ls emap bs bmap = fmap (\(LoanData loanId date personio_id iid) ->
-       Loan loanId (T.pack $ show date) (bookIdToItem bs bmap iid) (emap ^.at (P.EmployeeId (fromIntegral personio_id)))) ls
+    toItem binfos boainfos item =
+        case idInfoId item of
+          BookInfoId i      -> binfos ^.at i >>= (Just . Item (idItemId item) (idLibrary item) . ItemBook)
```
phadej: looks repetetive

phadej:
@neurogoo note, github doesn't notify when new commits are pushed to old PRs, so ping for re-review would be appreciated

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -19,152 +18,189 @@ import qualified Data.Map  as Map
...
+loans ctx = safePoolQuery ctx "SELECT loan_id, date_loaned, personio_id, item_id FROM library.loan" ()
 
 loan :: (MonadLog m, MonadBaseControl IO m, MonadCatch m) => Ctx -> LoanId -> m (Maybe LoanData)
-loan ctx lid = fmap listToMaybe $ safePoolQuery ctx "select id, date_loaned, book_id, personio_id from library_loan where id = ?" (Only lid)
-
-bookInformationToLoanable :: BookId -> BookInformation -> Library -> Loanable
-bookInformationToLoanable bid book = Loanable (Book bid book)
+loan ctx lid = listToMaybe <$> safePoolQuery ctx "SELECT loan_id, date_loaned, personio_id, item_id FROM library.loan where loan_id = ?" (Only lid)
 
 fetchLoan :: (MonadLog m, MonadBaseControl IO m, MonadCatch m) => Ctx -> LoanId -> IdMap P.Employee -> m (Maybe Loan)
```
phadej: swap `IdMap P.Employee` and `LoanId`, latter is varied more often.

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -19,152 +18,189 @@ import qualified Data.Map  as Map
...
-bookArrayToMap :: [BookData] -> Map BookId (BookInformationId, Library)
-bookArrayToMap = Map.fromList . fmap (\(BookData bid binfoid lib) -> (bid, (binfoid, lib)))
-
-bookIdToInformation :: BookId -> Map BookId (BookInformationId, Library) -> Map BookInformationId BookInformation -> Maybe BookInformation
-bookIdToInformation bid bookidmap bookinfomap = bookidmap ^.at bid >>= (\(infoid, _) -> bookinfomap ^.at infoid)
+      bookInformation item =
+          case idInfoId item of
+              BookInfoId info -> Just info
+              _               -> Nothing
+      loan' (LoanData loanId date personio_id itemId) emap bookInfo lib =
```
phadej: no need to rename `es` to `emap` as you closed over it anyway.

library-app/src/Futurice/App/Library/Logic.hs
phadej: (Another point: `IdMapP.Employee` is kind-of part of the context.

neurogoo: So would it be wise to add it to Ctx and add periodically updating job for it?

phadej: In short now. We want `IdMap` to be constant for the time we process the request. The `TVar (IdMap P.Employee)` could be in `Ctx` and periodically updated (so we don't spam `personio-proxy`), but that's different concern.

neurogoo: Hmm well isn't it currently constant for the duration of the request? As the data is fetched always once before going to the logic functions.

phadej: It is, but your ordering of arguments subtly implies it isn't. Change to

```haskell
... => Ctx -> IdMap P.Employee -> LoanId
```

then, multiple loans can be fetched with

```haskell
loans <- traverse (fetchLoan ctx employees) loanIds
```

neurogoo: Ah that makes sense :) Updating branch now

library-app/src/Futurice/App/Library/Types/Library.hs
```diff
@@ -50,5 +50,5 @@ instance FromJSON Library where
     parseJSON = withObject "Library" $ \l -> do
         office <- l .: "office"
         case officeFromText office of
-          Just o -> pure $ OfficeLibrary o
```
phadej: why object, and not simply a `String`? Can be changed later if there are no real reason.

PR #598, opened 2018-08-15
Preliminary library search and UI loan and borrow

neurogoo:


library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
```diff
@@ -0,0 +1,49 @@
...
+import Futurice.App.Library.Markup
+import Futurice.App.Library.Types
+
+import qualified Data.Text as T
+
+personalLoansPage :: [Loan] -> HtmlPage ("personalinformation")
+personalLoansPage loans = page_ "My loans" (Just NavUser) $ do
+    fullRow_ $ do
+        for_ (listToMaybe $ fst $ loanSorter loans) $ \_ -> do
+            h1_ $ "Books"
```
phadej: don't use `h1_`

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -202,17 +187,98 @@ loanDataToLoan ctx es ldatas = do
...
+                             (("SELECT bookinfo_id, title, isbn, author, publisher, publishedYear, cover, amazon_link FROM library.bookinformation " <> startPos criteria direction <> " AND " <> searchSqlString <> sortCriteria criteria direction) <> " LIMIT ?")
+                             $ (startInfo criteria info, info ^. bookInformationId, s, limit)
+      (Nothing, Nothing)  -> safePoolQuery ctx ("SELECT bookinfo_id, title, isbn, author, publisher, publishedYear, cover, amazon_link FROM library.bookinformation " <> sortCriteria criteria direction <> " LIMIT ?") (Only limit)
+      (Nothing, Just s)   -> safePoolQuery ctx ("SELECT bookinfo_id, title, isbn, author, publisher, publishedYear, cover, amazon_link FROM library.bookinformation WHERE" <> searchSqlString <> sortCriteria criteria direction <> " LIMIT ?") (s, limit)
+    where
+      searchSqlString = " bookinfo_id in (SELECT bookinfo_id FROM (SELECT bookinfo_id, title, to_tsvector(title) || to_tsvector(isbn) || to_tsvector(author) || to_tsvector(publisher) || to_tsvector(to_char(publishedYear, '9999')) as document FROM library.bookinformation) book_search WHERE book_search.document @@ to_tsquery(?)) "
```
phadej: this looks scary. Maybe we need to learn `opaleye` or `selda` at last :/

phadej: i.e. ok for now, but let's find a better solution soon.

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
```diff
@@ -0,0 +1,49 @@
...
+                        tr_ $ do
+                            td_ $ toHtml $ book ^. bookTitle
+                            td_ $ toHtml $ book ^. bookAuthor
+                            td_ $ toHtml $ book ^. bookPublisher
+                            td_ $ toHtml $ show $ book ^. bookPublished
+                            td_ $ toHtml $ book ^. bookISBN
+                            td_ $ button_ [class_ "button", data_ "futu-id" "return-loan", data_ "loan-id" (T.pack $ show lid)] $ toHtml ("Return" :: Text)
+        for_ (listToMaybe $ snd $ loanSorter loans) $ \_ -> do
+            h1_ $ "Boardgames"
+
```
phadej: ?

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
```diff
@@ -0,0 +1,49 @@
...
+
+  where
+    loanSorter :: [Loan] -> ([(LoanId, BookInformation)], [(LoanId, BoardGameInformation)])
+    loanSorter [] = ([],[])
+    loanSorter (l : ls) = case l ^. loanInformation ^. itemInfo of
```
phadej: could `partitionEithers` make this less manual?

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
```diff
@@ -0,0 +1,49 @@
...
+import Futurice.App.Library.API
+import Futurice.App.Library.Markup
+import Futurice.App.Library.Types
+
+import qualified Data.Text as T
+
+personalLoansPage :: [Loan] -> HtmlPage ("personalinformation")
+personalLoansPage loans = page_ "My loans" (Just NavUser) $ do
+    fullRow_ $ do
+        for_ (listToMaybe $ fst $ loanSorter loans) $ \_ -> do
```
phadej: Would  `unless (null $ fst $ loanSorter loans)` tell the intent better?

phadej: maybe even
```haskell
let bookLoans = fst $  loanSorter loans
unless (null bookLoans) $ do
    ...
    for_ bookLoans $ \(lid, book) -> ....
```

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
neurogoo: hmm, why is that? I do want h1 element there.

library-app/src/Futurice/App/Library/Logic.hs
neurogoo: Yeah, it felt a little dirty to write that but feels like db would be still right place to do that ordering/sorting/filtering when possible. I will add some typing to that and if that is not enough then look into those libraries

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
phadej: So far we had only one H1 element on a page, which would mirror the title of that page. In this case, it would be *My loans*. If the `page_` is from `Futurice.Lucid.Navigation`, it outputs that already: https://github.com/futurice/haskell-mega-repo/blob/master/futurice-foundation/src/Futurice/Lucid/Navigation.hs#L43-L50

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
neurogoo: A lot of things is still mostly book based as I have not yet had good ideas to implement some thing (like main page) for boardgames, but I will focus boardgames soon

phadej: leave TODOs, then we can omit them in the reviews (or pinpoint the forgotten ones).

library-app/src/Futurice/App/Library/Pages/PersonalLoansPage.hs
neurogoo: I was thinking about Either, but if we someday get third category then there would be a lot of refactoring. But do you think that it should still use for now Either?

phadej: still `Either`.

neurogoo: Hmm what would then be the update path when we get third category? Some kind of recursive Either or similar solution as the current one?

phadej: Don't worry about it now :)

neurogoo:
I tried typing the SQL, but I didn't really get satisfactory result, so I will revisit it later. So if other parts of the code are now fine, I think this should be merged

PR #602, opened 2018-08-30
Library add new item

neurogoo:
Page for adding new books and boardgames to database


library-app/src/Futurice/App/Library/Types/BookInformationMagicResponse.hs
```diff
@@ -0,0 +1,48 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE InstanceSigs      #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TypeFamilies      #-}
+{-# LANGUAGE TypeOperators     #-}
+module Futurice.App.Library.Types.BookInformationMagicResponse where
```
phadej: magic?

library-app/src/Futurice/App/Library/Types/Library.hs
```diff
@@ -25,6 +26,19 @@ data Library
...
+
+libraryToText :: Library -> Text
+libraryToText (OfficeLibrary office) = officeToText office
+libraryToText Elibrary = "Elibrary"
+libraryToText UnknownLibrary = "Unknown"
+
+libraryFromText :: Text -> Maybe Library
+libraryFromText library = case officeFromText library of
+    Just office -> pure $ OfficeLibrary office
+    Nothing -> if library == "Elibrary" then pure Elibrary else pure UnknownLibrary
```
phadej:
```haskell
Nothing | library == "ELibrary" -> pure ELibrary
        | otherwise             -> pure unknownLibrary
```

library-app/src/Futurice/App/Library/Types/BookInformationMagicResponse.hs
```diff
@@ -0,0 +1,48 @@
...
+                deriving (Eq, Ord, Show, Generic, ToJSON, ToSchema)
+
+data BooksPerLibrary = BooksPerLibrary
+    { _bookLibrary   :: !Library
+    , _amountOfBooks :: !Int
+    }
+    deriving (Eq, Ord, Show, Generic, ToSchema, Typeable)
+
+data BookInformationMagicResponse = BookInformationMagicResponse
+    { _idM              :: !BookInformationId
```
phadej: we should have some prefix here, rather than suffix.

library-app/src/Futurice/App/Library/Types/BookInformationMagicResponse.hs
```diff
@@ -0,0 +1,48 @@
...
+    , _titleM           :: !Text
+    , _ISBNM            :: !Text
+    , _authorM          :: !Text
+    , _publisherM       :: !Text
+    , _publishedM       :: !Int
+    , _coverM           :: !Text
+    , _amazonLinkM      :: !Text
+    , _booksM           :: ![BooksPerLibrary]
```
phadej: this can be `Map Library Int`

library-app/src/Futurice/App/Library/Types/BookInformationMagicResponse.hs
neurogoo: That was how this option was named in the old app :D But I will change that to ByISBN as that I have used elsewhere

library-app/src/Futurice/App/Library/Pages/AddItemPage.hs
```diff
@@ -0,0 +1,77 @@
...
+import Futurice.Prelude
+import Prelude ()
+
+import Futurice.App.Library.API
+import Futurice.App.Library.Markup
+import Futurice.App.Library.Types
+
+addItemPage :: HtmlPage "additempage"
+addItemPage = page_ "Add item to Library" (Just NavAddItem) $ do
+    div_ $ do
```
phadej: is this div needed?

library-app/src/Futurice/App/Library.hs
```diff
@@ -164,6 +178,16 @@ getBookImpl ctx lid = do
       Just info -> pure info
       Nothing -> throwError $ err404 { errBody = "No bookinformation found"}
 
+getBookByISBNImpl :: Ctx -> Text -> Handler BookInformationByISBNResponse
+getBookByISBNImpl ctx isbn = do
+    info <- runLogT "library" (ctxLogger ctx) $ fetchBookInformationByISBN ctx isbn
+    case info of
+      Just (BookInformationResponse infoid title _ author publisher published cover amazonLink books) ->
```
phadej: here I'd use record syntax. Those are text fields, so it's easy to swap them (think when one more is added).

neurogoo:
Didn't notice duplicate commit message ^^'

PR #603, opened 2018-09-05
Library use sisosota

neurogoo:
Some validations for form data and finally switching to use Sisosota as cover picture source. I

### DIFF
--- a/futurice-constants/src/Futurice/Constants.hs
+++ b/futurice-constants/src/Futurice/Constants.hs
@@ -10,7 +10,6 @@ module Futurice.Constants (
     planmillPublicUrl,
     planmillPublicUrlStr,
     servicePublicUrl,
-    sisosotaPublicUrl,
     supportEmailHtml,
     competenceMap,
     ) where
@@ -60,9 +59,6 @@ planmillPublicUrl = servicePublicUrl PlanmillService
 
 planmillPublicUrlStr :: String
 planmillPublicUrlStr = planmillPublicUrl ^. unpacked
-
-sisosotaPublicUrl :: Text
-sisosotaPublicUrl = servicePublicUrl SisosotaService
 
 -------------------------------------------------------------------------------
 -- Other constants
--- a/futurice-constants/src/Futurice/Constants.hs
+++ b/futurice-constants/src/Futurice/Constants.hs
@@ -10,6 +10,7 @@ module Futurice.Constants (
     planmillPublicUrl,
     planmillPublicUrlStr,
     servicePublicUrl,
+    sisosotaPublicUrl,
     supportEmailHtml,
     competenceMap,
     ) where
@@ -59,6 +60,9 @@ planmillPublicUrl = servicePublicUrl PlanmillService
 
 planmillPublicUrlStr :: String
 planmillPublicUrlStr = planmillPublicUrl ^. unpacked
+
+sisosotaPublicUrl :: Text
+sisosotaPublicUrl = servicePublicUrl SisosotaService
 
 -------------------------------------------------------------------------------
 -- Other constants
--- a/library-app/library-app.js
+++ b/library-app/library-app.js
@@ -36,14 +36,12 @@ futu.onload(function () {
             s.remove();
         });
         var isbn = $("form[data-futu-id=add-new-book] input[name='isbn']").value;
-        for(var i=0; i < isbn.length; i++) {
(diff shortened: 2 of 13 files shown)


futurice-constants/src/Futurice/Constants.hs
```diff
@@ -60,6 +61,9 @@ planmillPublicUrl = servicePublicUrl PlanmillService
 planmillPublicUrlStr :: String
 planmillPublicUrlStr = planmillPublicUrl ^. unpacked
 
+sisosotaPublicUrl :: Text
```
phadej: don't add this, `servicePublicUrl SisosotaService` is short enough on itself.

library-app/library-app.js
```diff
@@ -31,6 +31,23 @@ futu.onload(function () {
     });
   });
 
+    $("form[data-futu-id=add-new-book]").onsubmit = function () {
+        $$('span.alert').forEach(function (s) {
+            s.remove();
+        });
+        var isbn = $("form[data-futu-id=add-new-book] input[name='isbn']").value;
+        for(var i=0; i < isbn.length; i++) {
```
phadej: won't a regexp match be easier for filtering non-numbers?

library-app/src/Futurice/App/Library.hs
```diff
@@ -96,6 +100,19 @@ getPersonioDataMap ctx = do
     pure $ Map.fromList $ catMaybes $ fmap (\e -> case e ^. P.employeeLogin of
                                                Just login -> Just (login, e)
                                                Nothing -> Nothing) es'
+
+fetchCover :: (Monad m, MonadIO m, MonadThrow m) => Ctx -> Text -> m (Either String DBL.ByteString)
```
phadej: can't this function take `ContentHash`?

library-app/src/Futurice/App/Library.hs
```diff
@@ -107,12 +124,34 @@ getPersonioDataMap ctx = do
...
+_updateAllBookCovers ctx = do
+    books <-  runLogT "library" (ctxLogger ctx) $ fetchBookInformations ctx
+    for_ books fetchAndSendCover
+  where
+      fetchAndSendCover :: BookInformation -> Handler ()
+      fetchAndSendCover bookInformation = do
+          pictData <- if (all isSpace (T.unpack $ bookInformation ^. bookCover)) || not (T.isPrefixOf "cover" (bookInformation ^. bookCover)) then
+                        pure Nothing
+                      else
+                        Just <$> liftIO (DBL.readFile (T.unpack $ "/Users/toku/hmr/library-app/media/" <> (bookInformation ^. bookCover)))
```
phadej: `/users/toku`....

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -290,25 +293,28 @@ bookIdToInformation iid bookidmap bookinfomap = do
...
         bookInformation <- MaybeT $ fetchBookInformation ctx info
         if bookInformation ^. bookISBN == isbn then pure $ bookInformation ^. bookInformationId else MaybeT $ pure Nothing
     bookInfoId <- case checkedInfoId of
                     Nothing ->  listToMaybe <$> safePoolQuery
                                 ctx
-                                -- TODO add cover link
-                                "INSERT INTO library.bookinformation (title, isbn, author, publisher, publishedyear, amazon_link) VALUES (?,?,?,?,?,?) returning bookinfo_id"
-                                (title, isbn, author, publisher, published, amazonLink)
+                                "INSERT INTO library.bookinformation (title, isbn, author, publisher, publishedyear, cover, amazon_link) VALUES (?,?,?,?,?,?,?) returning bookinfo_id"
+                                (title, isbn, author, publisher, published, contentHashToText contentHash, amazonLink)
```
phadej: rather add `Postgres.ToField ContentHash` and `Postgres.FromField ContentHash` instances if there aren't those.

futurice-constants/src/Futurice/Constants.hs
neurogoo: I will remove this, but isn't this true with every other service url in this file? ^^'

phadej: Yes. They were there before `Service` data type were introduced, but in the long run, they are better removed all together.

library-app/library-app.js
neurogoo: Well the old adage is that with regexp you have two problems instead of one so I rarely use it unless it is necessary 😅 But you are right, here that shouldn't be a problem

PR #608, opened 2018-09-19
Library edit information 

neurogoo:
Add boardgame support to all relevant pages and add way to edit information


library-app/library-app.js
```diff
@@ -31,22 +31,48 @@ futu.onload(function () {
...
+    });
+
+    $$("form[data-futu-id=add-new-book]").forEach(function (frm) {
+        frm.onsubmit = function () {
+            $$('span.alert').forEach(function (s) {
+                s.remove();
+            });
+            var isbn = $("form[data-form-type=book] input[name='isbn']").value;
+            var noErrors = true;
+            noErrors = isbnCheck(isbn);
```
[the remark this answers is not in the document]
phadej: or return immediately if `false`, as in

```haskell
if (!isbnCheck(isbn)) return false;
```

library-app/src/Futurice/App/Library/Pages/BoardGameInformationPage.hs
```diff
@@ -0,0 +1,67 @@
...
+import qualified Personio  as P
+
+boardGameInformationPage :: BoardGameInformationResponse -> [LoanData] -> IdMap P.Employee -> HtmlPage ("boardgameinformation")
+boardGameInformationPage boardGameResponse ls employees = page_ "Board Game details" (Nothing :: Maybe Nav) $ do
+    fullRow_ $ do
+        div_ [] $ do
+            a_ [class_ "button small", href_ $ linkToText $ fieldLink editBoardGamePageGet (boardGameResponse ^. boardGameResponseInformationId)] "Edit board game information"
+            table_ $ do
+                tr_ $ do
+                    th_ "Name"
```
phadej: we have `vertRow` in `Futurice.Lucid.Foundation`

```haskell
vertRow_ :: Monad m => Text -> HtmlT m () -> HtmlT m ()
vertRow_ title h = tr_ $ do
    th_ [ style_ "white-space: nowrap" ] (toHtml title)
    td_ [ style_ "width: 100%" ] h
```

library-app/src/Futurice/App/Library/Pages/BookInformationPage.hs
```diff
@@ -52,7 +53,7 @@ bookInformationPage (BookInformationResponse binfoid title isbn author publisher
                                data_ "futu-id" "loan-item",
                                data_ "item-id" (T.pack $ show binfoid),
                                data_ "library" (T.pack $ show $ toHtml library)] $ toHtml ("Borrow" :: Text)
-                  table_ $ do
+                  for_ (listToMaybe $ fst (partitionByLoan bs)) $ \_ -> table_ $ do
```
phadej: unless (null ...) ?

library-app/library-app.js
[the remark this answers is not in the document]
neurogoo: I don't want to return immediately as I want to inform about all the errors at the same time. But I should assign immediately

library-app/src/Futurice/App/Library.hs
```diff
@@ -355,14 +377,38 @@ addBookPostImpl ctx addBook@AddBookInformation{..} = do
...
 addBoardGamePostImpl ctx addBoardGame = do
     res <- runLogT "library" (ctxLogger ctx) $ addNewBoardGame ctx addBoardGame
     if res then
       addItemPageGetImpl ctx
     else
       throwError err400
 
+editBookInformationPageImpl :: Ctx -> BookInformationId -> Handler (HtmlPage "edititempage")
+editBookInformationPageImpl ctx infoId = do
+    info <- runLogT "library" (ctxLogger ctx) $ fetchBookInformation ctx infoId
```
phadej: I'd use "command" name for logging. `library` part is obvious already.

library-app/src/Futurice/App/Library/Logic.hs
```diff
@@ -188,72 +188,95 @@ loanDataToLoan ctx es ldatas = do
...
+startBookInfo SortAuthor info = info ^. bookAuthor
+startBookInfo SortISBN info = info ^. bookISBN
+startBookInfo SortPublished info = T.pack $ show $ info ^. bookPublished
+
+startBoardGameInfo :: BoardGameSortCriteria -> BoardGameInformation -> Text
+startBoardGameInfo SortName info = info ^. boardGameName
+startBoardGameInfo SortDesigner info = fromMaybe "" $ info ^. boardGameDesigner
+
+searchSqlString :: SortCriteria -> Query
+searchSqlString (BookSort _) = " bookinfo_id in (SELECT bookinfo_id FROM (SELECT bookinfo_id, title, to_tsvector(title) || to_tsvector(isbn) || to_tsvector(author) || to_tsvector(publisher) || to_tsvector(to_char(publishedYear, '9999')) as document FROM library.bookinformation) book_search WHERE book_search.document @@ plainto_tsquery(?)) "
```
phadej: This query mess starts to be so complicated, that we need to setup tests testing them.

Testing at least `SELECT` for syntax correctness (i.e. not care what db returns) would be easy. Travis has postgres.

phadej: an alternative is to use e.g. http://hackage.haskell.org/package/selda

neurogoo: I do wonder if Selda support Postgres full text search. Browsing through docs I at least don't find any mention of that.

phadej: then we have to write tests.

phadej:
```diff
-testConnection :: (MonadIO m) => m (Pool Connection)
-testConnection = createPostgresPool defaultConnectInfo
+testConnection :: (MonadIO m, MonadLog m) => m (Pool Connection)
+testConnection = do
+    connectInfo <- getConfig' "TEST" envConnectInfo
+    createPostgresPool connectInfo
```

---

For tests: don't use stderr-logger, but create a logger which has a state, collecting whether there were `ATTENTION` messages or not. Then the test runner can check whether there were, and print the error message.

library-app/tests/Tests.hs
```diff
@@ -0,0 +1,57 @@
...
+    , testCase "Search boardgameInformation" $ withStderrLogger $ \logger -> do
+          pp <- runLogT "testing" logger testConnection
+          for_ allBoardGameSortCriteriaAndStart $ \cri -> for allDirections $ \dir ->
+            runLogT "testing" logger $ fetchInformationsWithCriteria pp cri dir testLimit Nothing :: IO [BoardGameInformation]
+          pure ()
+    , testCase "Search boardgameInformation with search" $ withStderrLogger $ \logger -> do
+          pp <- runLogT "testing" logger testConnection
+          for_ allBoardGameSortCriteriaAndStart $ \cri -> for allDirections $ \dir ->
+            runLogT "testing" logger $ fetchInformationsWithCriteria pp cri dir testLimit (Just "testing") :: IO [BoardGameInformation]
+          pure ()
```
phadej: pure is unneeded

library-app/tests/Tests.hs
```diff
@@ -0,0 +1,57 @@
...
+    , testCase "Search boardgameInformation" $ withStderrLogger $ \logger -> do
+          pp <- runLogT "testing" logger testConnection
+          for_ allBoardGameSortCriteriaAndStart $ \cri -> for allDirections $ \dir ->
+            runLogT "testing" logger $ fetchInformationsWithCriteria pp cri dir testLimit Nothing :: IO [BoardGameInformation]
+          pure ()
+    , testCase "Search boardgameInformation with search" $ withStderrLogger $ \logger -> do
+          pp <- runLogT "testing" logger testConnection
+          for_ allBoardGameSortCriteriaAndStart $ \cri -> for allDirections $ \dir ->
+            runLogT "testing" logger $ fetchInformationsWithCriteria pp cri dir testLimit (Just "testing") :: IO [BoardGameInformation]
```
phadej: runLogT can be done once per test

```haskell
withTestLogger $  \loggerState logger -> do
    runLogT $ ...
    assertNoAttentionLogMessages loggerState
```

neurogoo: Is assertNoAttentionLogMessages an actual function? ^^'

phadej: yes, after you write it

phadej: amendum: you need to write `wiithTestLogger` too, see other comments

phadej: We are only interested in Attention messages. So it's almost what we need, but not exactly.

phadej:
merged as #612

PR #613, opened 2018-10-03
Library add more filtering + forced loan

neurogoo:
Add way to filter based on library and availability and add way to do snatch aka forced loan on item

### DIFF
--- a/library-app/src/Futurice/App/Library.hs
+++ b/library-app/src/Futurice/App/Library.hs
@@ -249,11 +249,12 @@ indexPageImpl :: Ctx
               -> Maybe BookInformationId
               -> Maybe BoardGameInformationId
               -> Maybe Text
-              -> Maybe HttpApiDataLibrary
+              -> Maybe LibraryOrAll
               -> Maybe Text
               -> Handler (HtmlPage "indexpage")
 indexPageImpl ctx criteria direction limit startBookId startBoardGameId search library onlyAvailableText = do
-    let lib = library >>= libraryData
+    -- default values
+    let lib = fromMaybe AllLibraries library
         onlyAvailable = isJust onlyAvailableText
         cleanedSearch = search >>= (\s -> if s == "" then Nothing else Just s)
         crit = fromMaybe (BookSort SortTitle) criteria
@@ -262,7 +263,6 @@ indexPageImpl ctx criteria direction limit startBookId startBoardGameId search l
     case crit of
       (BookSort bookCrit) -> do
           bookInfo <- maybe (pure Nothing) (\x -> runLogT "fetch-information" (ctxLogger ctx) $ fetchBookInformation (ctxPostgres ctx) x) startBookId
-          _ <- runLogT "test" (ctxLogger ctx) $ logInfo_ $ "Only available is " <> T.pack (show onlyAvailable)
           bookInfos <- runLogT "fetch-information-with-criteria" (ctxLogger ctx) $ fetchInformationsWithCriteria (ctxPostgres ctx) (BookCS bookCrit bookInfo) dir lim cleanedSearch lib onlyAvailable
           pure $ indexPage (BookCD bookCrit bookInfos) dir lim startBookId startBoardGameId search library onlyAvailableText
       (BoardGameSort boardgameCrit) -> do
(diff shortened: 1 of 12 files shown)


library-app/src/Futurice/App/Library.hs
```diff
@@ -299,16 +306,19 @@ borrowBookImpl ctx login req = withAuthUser ctx login $ (\l -> do
...
     emap <- liftIO $ getPersonioDataMap ctx
-    case emap ^.at l of
-      Nothing -> throwError err403
-      Just es -> do
-          loanData <- runLogT "snatch-book" (ctxLogger ctx) $ snatchBook (ctxPostgres ctx) (es ^. P.employeeId) iid
-          case loanData of
-            Nothing -> throwError $ err404 { errBody = "Couln't loan book" }
-            Just (LoanData lid _ _ _) -> getLoanImpl ctx lid)
+    newLoan <- runMaybeT $ do
+        loanData <- MaybeT $ runLogT "fetch-loanid" (ctxLogger ctx) $ fetchLoanIdWithItemId (ctxPostgres ctx) iid
```
phadej: try to use single `runLogT`

library-app/src/Futurice/App/Library/API.hs
```diff
@@ -63,6 +63,8 @@ type IndexPageEndpoint = QueryParam "criteria" SortCriteria
                          :> QueryParam "start-book" BookInformationId
                          :> QueryParam "start-boardgame" BoardGameInformationId
                          :> QueryParam "search" Text
+                         :> QueryParam "library" HttpApiDataLibrary
```
phadej: why this cannot be just `Library`? `QueryParam` is optional already...

library-app/src/Futurice/App/Library/Types/AddItemRequest.hs
```diff
@@ -46,7 +46,12 @@ data AddBoardGameInformation = AddBoardGameInformation
     deriving Show
 
 usedLibraries :: [Library]
-usedLibraries = (filter (/= OfficeLibrary offOther) . filter (/= UnknownLibrary)) allLibraries
+usedLibraries = sortBy (\a b -> case (a,b) of
```
phadej: show this order differes from `Ord`. Make it as a named function

library-app/src/Futurice/App/Library/Types/Loan.hs
```diff
@@ -51,6 +52,8 @@ instance ToJSON Loan where
         , "library" .= _itemLibrary item
         ]
 
+instance FromRow LoanId where
```
phadej: Let's not do this, but use `Only`. This may cause accidental errors.

library-app/src/Futurice/App/Library/API.hs
neurogoo: Problem was that select box send always some input, so it can't be just Library. And Maybe Library felt still best as All Libraries == no library selected

library-app/src/Futurice/App/Library.hs
neurogoo: I was considering this, but I was thinking that there is value of having different runLogT to each to have different prefix for helping to debug in case of problems. But you think that that doesn't offset the redundancy of that code?

library-app/src/Futurice/App/Library/Types/AddItemRequest.hs
neurogoo: something like `sortBy librarySelectSortOrder $ (filter (/= OfficeLibrary offOther) . filter (/= UnknownLibrary)) allLibraries` ?

phadej: yeah

library-app/src/Futurice/App/Library.hs
phadej: error message prints the query too, if you worry about that.

also http://hackage.haskell.org/package/log-base-0.7.4.0/docs/Log-Class.html#v:localDomain if you really want to change prefixes

phadej: i.e. on top level we should have `runLogT "snatch" lgr $  do ... ` and individual commands may use `localDomain`, but I don't think that's needed.

PR #615, opened 2018-10-16
Timereports Kind

phadej:


hours-api/src/Futurice/App/HoursApi/Logic.hs
```diff
@@ -278,14 +279,24 @@ userResponse = User
...
+            TK.KindSickLeave      -> mempty
+            TK.KindBalanceAbsence -> mempty
           where
             NDT hours = fmap realToFrac (report ^. H.timereportAmount) :: NDT 'Hours Float
 
 entryUpdateResponse :: H.MonadHours m => Day -> m EntryUpdateResponse
 entryUpdateResponse d =
     EntryUpdateResponse <$> userResponse <*> hoursResponse (d ... d)
+
+kindToType :: TK.TimereportKind -> EntryType
```
neurogoo: I'm a little confused on why to create kind type that is structurally identical to EntryType? Is it just to separete completely Hours from Planmill

phadej: it's not structurally identical. `TimereportKind` has more variants.

futurice-integrations/src/Futurice/Integrations/TimereportKind.hs
```diff
@@ -0,0 +1,65 @@
...
+    dutyType <- traverse (`PMQ.enumerationValue` "D") (PM.trDutyType tr)
+
+    pid <- case PM.trProject tr of
+        Just pid -> return (Just pid)
+        Nothing  -> PM.taskProject <$> PMQ.task (PM.trTask tr)
+
+    mproject <- traverse PMQ.project pid
+    maccount <- traverse PMQ.account (mproject >>= PM.pAccount)
+
+    mprojectCategory <- traverse (`PMQ.enumerationValue` "C?") (PM.pCategory <$> mproject)
```
neurogoo: What is "C?" or "T?"? If I understand correctly, it is defaultvalue, but feels like a weird value.

phadej: just random strings, different from each other. (helps debugging)

phadej: C = category, T = type

phadej: ... In the longer run we could make `enumerationValue` return `Maybe`, which will be better, but not now.

PR #635, opened 2018-11-19
Futurice.Postgres.SqlBuilder

phadej:


phadej:
`selda` looks good, but unfortunately doesn't build on top of `postgresql-simple`. So migration would be very painful. (I.e. you have to rewrite all the queries).

PR #640, opened 2018-11-22
Improve UI

neurogoo:


neurogoo:
Search didn't work as title wasn't send right after the indexPage input type change

PR #649, opened 2018-12-17
Ghc 8.6 only

phadej:


cabal.project
```diff
@@ -1,5 +1,6 @@
 index-state: 2018-12-04T11:13:52Z
 tests: True
+with-compiler: ghc-8.6.3
```
phadej: Note: that can still be overriden with cli flag, if needed. (e.g. in travis)

PR #651, opened 2018-12-19
Futuqu accounts

phadej:


phadej:
using API.Generic didn't speed up the compilation, but at least I removed some cruft

reports-app/src/Futurice/App/Reports.hs
```diff
@@ -317,8 +303,8 @@ defaultMain = futuriceServerMain (const makeCtx) $ emptyServerConfig
         let missingHoursStatsJob =
                 mkJob "missing-hours-stats" (missingHoursStats ctx) $ every (60 * 60)
 
-        let jobs = missingHoursStatsJob : hcollapse
```
neurogoo: What did this do and why remove it?

phadej: it kept reports "ready", but they are fast to regenerate; so it doesn't really buy anything

PR #659, opened 2019-01-16
New Power endpoints

mixman:


proxy-types/src/Futurice/App/Proxy/API.hs
```diff
@@ -94,7 +94,36 @@ data Routes = Routes
     , routeFutuquTimereportsStream :: Futuqu ("strm" :> "timereports.csv.gz" :> QueryParam "since-month" Month :> StreamGet NoFraming (CACHED (GZIP CSV)) (SourceIO (Cached (GZIP CSV) Text)))
     , routeFutuquCapacitiesStream  :: Futuqu ("strm" :> "capacities.csv.gz"  :> QueryParam "since-month" Month :> StreamGet NoFraming (CACHED (GZIP CSV)) (SourceIO (Cached (GZIP CSV) Text)))
 
-    -- Power
+    -- Power v3
+    , routePowerV3PersonCompetences :: ProxiedEndpoint 'PowerService
+        ("uiapi" :> "timeline" :> "personcompetence/" :> Get '[JSON] Value)
+        ("power" :> "v3api" :> "person_competences" :> Get '[JSON] Value)
```
neurogoo: I would probably do the public urls as for example `"power" :> "v3" :> "api" :> "projects"` as that kind of format is more familiar on the outside world (and was used in old Power)

PR #663, opened 2019-01-24
Feature/proxy mgmt stats

jarcane:


proxy-mgmt-app/src/Futurice/App/ProxyMgmt/Pages/Tokens.hs
```diff
@@ -71,15 +77,29 @@ tokensPage policies tokens aes = page_ "Audit log" (Just NavTokens) $ do
...
+    makeCounts :: [AccessEntry] -> (LenientEndpoint, Int)
+    makeCounts xs = (aeEndpoint $ head xs, length xs)
+
 -------------------------------------------------------------------------------
 -- Util: move to futurice-prelude
 -------------------------------------------------------------------------------
 
+eqBy :: Eq b => (a -> b) -> a -> a -> Bool
+eqBy f x y = f x == f y 
+
```
neurogoo: I would use `comparing` function from Data.Ord

proxy-mgmt-app/src/Futurice/App/ProxyMgmt/Pages/Tokens.hs
```diff
@@ -71,15 +77,29 @@ tokensPage policies tokens aes = page_ "Audit log" (Just NavTokens) $ do
                 toHtml e
                 " → "
                 toHtml (formatHumanHelsinkiTime t)
+            let counts = map makeCounts . groupBy (eqBy aeEndpoint) $ filter isCurrMo ae
```
phadej: Free advice, `groupBy` is tricky, from `group` docs:

> The group function takes a list and returns a list of lists such that the concatenation of the result is equal to the argument. Moreover, each sublist in the result contains only equal elements. For example,

```haskell
>>> group "Mississippi"
["M","i","ss","i","ss","i","pp","i"]
```

---

```haskell
λ> :m +Data.List
λ Data.List> :m +Data.Function
λ D.List D.Function> import Data.Map as Map
λ D.List D.Function Map> groupBy ((==) `on` fst) [("a", "A1"), ("b", "B1"), ("b", "B2"), ("a", "A2")]
[[("a","A1")],[("b","B1"),("b","B2")],[("a","A2")]]
λ D.List D.Function Map> Map.fromListWith (+) [ (fst x, 1) | x <- [("a", "A1"), ("b", "B1"), ("b", "B2"), ("a", "A2")] ]
fromList [("a",2),("b",2)]
```

phadej: In fact, that's the trickery which

```haskell
calaf (UnionWith' . fmap Max) foldMap (\x -> Map.singleton (aeEndpoint x) (aeStamp x)) ae
```
above does.

You can write (for counting)

```haskell
calaf (UnionWith' . fmap Sum) foldMap (\x -> Map.singleton (aeEndpoint x) 1) ae
```

or (for max stamp)

```haskell
Map.fromListWith max [ (aeEndpoint x, aeStamp x) | x <- ae ]
```

---

I'm sorry I was excited with `alaf` :)

jarcane: I suspected there might be a way to use that, but I'll be perfectly honest: I didn't really understand it. 😂

It really doesn't help I don't have *any* type tooltips in my editor right now.

phadej: Paraphrasing SPJ:

> Trying to understand the type of `calaf` directly can lead to brain damage. It is easier to see what the instances look like.

```haskell
>>> alaf Sum foldMap Prelude.length ["hello","world"]
10

-- type of foldMap let us compose them:
>>> alaf Sum (foldMap . foldMap) length [["foo", "bar"], ["hello", "world"]]
16

-- alaf can be used with other (a -> b) -> c -> d looking functions:
>>> alaf ZipList traverse (\a -> [a,a]) "foo"
["foo","foo"]

-- i.e. we used different Applicative structure on List, default is:
>>> traverse (\a -> [a,a]) "foo"
["foo","foo","foo","foo","foo","foo","foo","foo"]
```

See more in http://hackage.haskell.org/package/newtype-0.2/docs/Control-Newtype.html (alaf = ala')

proxy-mgmt-app/src/Futurice/App/ProxyMgmt/Ctx.hs
```diff
@@ -10,11 +11,19 @@ import Futurice.Servant              (Cache)
...
+    , ctxDashdoServer :: !(Server DashdoAPI)
     }
+
+data DashdoCtx = DashdoCtx
+    { dctxPostgresPool :: !(Pool Connection)
+    , dctxConfig       :: !Config
+    , dctxLogger       :: !Logger
+    , dctxCache        :: !Cache
+    , dctxManager      :: !Manager
+    }
```
neurogoo: Why have separate Ctx for Dashdo?

jarcane: Because creating the Dashdo server requires Ctx, but Ctx requires Dashdo.

The previous code got around this circular dependency by wrapping the ctxDashdoServer field with the Identity functor and using Proxy as a filler until it can be properly populated, but this makes it messy to deal with later, and seems like overkill when all we're really trying to do is get these values to the function that generates the Dashdo server.

It's kind repetitive in its own right and there's possibly a more elegant solution I'm missing, but it's similar to the solution that reports-app uses for the same problem (there it just makes a tuple instead of a named type)

proxy-mgmt-app/src/Futurice/App/ProxyMgmt/Pages/Reports.hs
```diff
@@ -71,25 +71,29 @@ pieChart title field currMonth accessEntries =
...
 chartPerDay currMonth accessEntries = Chart $ C.toRenderable layout
     where
+        values' :: [(Day, [Double])]
         values' = map (\(s, v) -> (s, [fromIntegral v])) . assocs $ countByF currMonth (utctDay . aeStamp) accessEntries
 
-        values :: [(Int, [Double])]
-        values = map (\(s, v) -> ((\(_,_,t) -> t) $ toGregorian s, v)) values'
-        
-        titles = map (show . fst) values'
+        empty = Map.fromList [ (x, [0.0]) | x <- [1..31] ]
```
phadej: not every months has 31 days...

PR #670, opened 2019-02-05
No longer send notifications of missing hours to employees with end date

jarcane:
This should prevent the notification runner from sending messages to ex employees. 

This basically adds a check only on the notifications runner to not send notifications to someone with a set end date.

### DIFF
--- a/reports-app/src/Futurice/App/Reports/MissingHours.hs
+++ b/reports-app/src/Futurice/App/Reports/MissingHours.hs
@@ -64,14 +64,8 @@ missingHoursEmployeeNotificationPredicate :: Interval Day -> P.Employee -> Bool
 missingHoursEmployeeNotificationPredicate interval p = and
     [ p ^. P.employeeEmploymentType == Just P.Internal
     , p ^. P.employeeSalaryType == Just P.Monthly
-    , p `isAfter` sup interval
-    , P.employeeIsActiveInterval interval p
+    , P.employeeIsActiveWholeInterval interval p
     ]
-    where
-        isAfter :: P.HasSimpleEmployee e => e -> Day -> Bool
-        isAfter e d = case e ^. P.employeeEndDate of
-            Nothing -> True
-            Just ed -> ed > d
 
 
 -------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


reports-app/src/Futurice/App/Reports/MissingHours.hs
```diff
@@ -59,6 +60,14 @@ missingHoursEmployeePredicate interval p = and
     , P.employeeIsActiveInterval interval p
     ]
 
+missingHoursEmployeeNotificationPredicate :: Interval Day -> P.Employee -> Bool
+missingHoursEmployeeNotificationPredicate interval p = and
+    [ p ^. P.employeeEmploymentType == Just P.Internal
+    , p ^. P.employeeSalaryType == Just P.Monthly
+    , p ^. P.employeeEndDate == Nothing
```
neurogoo: Probably better to check that endDate < currentday as contract end date can be many months away(for example summer workers)

jarcane:
> 
> 
> Actually, this probably does not work correctly if employeeEndDate is Nothing?

Yeah, it seems you're correct. `Nothing` is always less than `Just` when comparing. So permanent employees would always return `False` and we'd never send anyone anything ... 😝

reports-app/src/Futurice/App/Reports/MissingHours.hs
```diff
@@ -64,9 +64,15 @@ missingHoursEmployeeNotificationPredicate :: Interval Day -> P.Employee -> Bool
 missingHoursEmployeeNotificationPredicate interval p = and
     [ p ^. P.employeeEmploymentType == Just P.Internal
     , p ^. P.employeeSalaryType == Just P.Monthly
-    , p ^. P.employeeEndDate > Just (sup interval)
+    , p `isAfter` sup interval
     , P.employeeIsActiveInterval interval p
     ]
+    where
+        isAfter :: P.HasSimpleEmployee e => e -> Day -> Bool
```
phadej: FWIW, this is so common, you might want to add a helper to `personio-client`.

E.g. grep for `maybe False` and `maybe True`

phadej: And thanks for fixing this ;)

jarcane: Huh, looking at it, it appears there is actually already a function for this specific use case in personio-client. https://github.com/futurice/haskell-mega-repo/blob/958b40bba9a990f6385dce75ffee346c92546dcf/personio-client/src/Personio/Types/SimpleEmployee.hs#L88

jarcane: Wait ... Also I think this is what the `employeeIsActiveInterval` function is supposed to be doing already, it just doesn't seem to work?

jarcane: And actually, it looks like `employeeIsActiveWholeInterval` does what we want for the notifications already, so wheee. :rofl:

PR #674, opened 2019-02-20
Split Germany into Tribes based on office

mixman:


reports-app/src/Futurice/App/Reports/PowerUser.hs
```diff
@@ -80,11 +81,12 @@ powerUserReport = do
 powerUser :: Day -> EmployeeMap -> P.Employee -> Maybe PowerUser
 powerUser today es e = do
     login <- e ^. P.employeeLogin
+    virtualTeam <- team
```
neurogoo: Why monadic function? Team function always returns Just a so no need for that

reports-app/src/Futurice/App/Reports/PowerUser.hs
```diff
@@ -100,6 +102,20 @@ powerUser today es e = do
...
+        then
+            if oname == "Munich"
+            then pure "Munich"
+            else
+                if oname == "Berlin"
+                then pure "Berlin"
+                else pure tname
+        else
+            pure tname
+
```
neurogoo: I would prefer something like this for example
```haskell
case (tname, oname) of
  ("Germany", "Munich") -> "Munich"
  ("Germany", "Berlin") -> "Berlin"
  _ -> tname
```

PR #676, opened 2019-02-25
Don't send missing hours notification on empty phone number

jarcane:


reports-app/src/Futurice/App/Reports/MissingHoursNotifications.hs
```diff
@@ -102,6 +102,7 @@ missingHoursNotifications ctx = runLogT "missing-hours-notifications" lgr $ do
 
                         when (pref ^. prefHoursPingSMS) $ case p ^. P.employeeWorkPhone of
                             Nothing -> logAttention "Employee without phone" login
+                            Just "" -> logAttention "Employee without phone" login
```
phadej: I'd rather make `_employeeWorkPhone` field `Nothing` if the input value is empty string. Fix the cause, not a symptom

neurogoo: Good point. I'll add ticket for that

jarcane: Yeah, this was a point I was thinking as well, and why I've not merged yet. Still looking if there's some way to maybe just fix the parsing in the first place so this shouldn't happen.

phadej: See e.g. https://github.com/futurice/haskell-mega-repo/blob/85e164a792d45165e0e7c4dba0a678da022b3a95/personio-client/src/Personio/Types/Employee.hs#L170

personio-client/src/Personio/Types/Employee.hs
```diff
@@ -194,5 +194,8 @@ parseEmployeeObject obj' = Employee
         | n <= 0    = Nothing
         | otherwise = Just n
 
+    notEmpty s | null s = Nothing
+    notEmpty s          = s
```
neurogoo: Shouldn't this be Just s? And first line could be shortened to notEmpty "" = Nothing. You could also strip empty spaces before checking for fancy points :D

neurogoo: This typechecks, but null (Just "") = False

jarcane: Ahh, yes, you're right. I was using the wrong `null` function and somehow it still type checked.

PR #677, opened 2019-02-26
Update stack.yaml to work with HIE

jarcane:
Attempting to do what I can to get this building again in Stack at least far enough to get one of either Intero or HIE to work.

So far I have had limited success. There seems to be an issue getting `stack ghci` to work, which is needed for Intero to work. I get an error that seems to have to do with the way avatar-app is setup: `<command line>: cannot satisfy -package z-avatar-app-z-avatar-internal`

HIE meanwhile has some kind of bug when using it with stack; I've commented on an issue for it here: https://github.com/haskell/haskell-ide-engine/issues/872  Using it without stack is supposed to be possible, but it instead fails because it can't find the system GHC for some reason, and it tries to use the old style v1 cabal commands which I suspect won't work ever anyway since we're using v2.


jarcane:
Fun new problem: 

```
/home/annaia/projects/haskell-mega-repo/flowdock-rest/src/Chat/Flowdock/REST/Internal.hs:22:1: error:
    Ambiguous module name ‘Prelude.Compat’:
      it was found in multiple packages:
      base-compat-0.10.5 base-compat-batteries-0.10.5
   |
22 | import Prelude.Compat
   | ^^^^^^^^^^^^^^^^^^^^^
Failed, one module loaded.
```

flowdock-rest does not actually try to use both packages at once, so the normal Cabal build works fine, but since they're both in the global stack config, Intero/Stack/GHCi get confused trying to resolve the conflict and build fails and thus so does Intero (but only after spinning my cores for 10+ minutes trying to resolve it.)

jarcane:
It works!

With this branch and the newest version of [haskell-ide-engine](https://github.com/haskell/haskell-ide-engine#installation), the tools all seem to work as expected mostly.

There's still a few quirks: it seems to have difficulty doing "Go to definition" across submodules, but you still get cross-module tooltips so it's easy enough to follow those to their sources. Alternately, you could install `hasktags` to give it a little help.

Since Intero apparently has performance issues with large files, this seems the best way to go.

To install HIE, use the cabal-based installation, partly for compatibility and partly because it's about a million times faster:

```
git clone https://github.com/haskell/haskell-ide-engine --recurse-submodules
cd haskell-ide-engine
stack install.hs cabal-hie-8.6.3
stack install.hs cabal-build-doc-8.6.3
```

futurice-chart/futurice-chart.cabal
```diff
@@ -9,7 +9,6 @@ bug-reports:   https://github.com/futurice/haskell-mega-repo/issues
 author:        Oleg Grenrus <oleg.grenrus@iki.fi>
 maintainer:    Oleg Grenrus <oleg.grenrus@iki.fi>
 license:       BSD-3-Clause
-license-file:  LICENSE
 build-type:    Simple
```
neurogoo: Rather than remove, I would add the LICENCE files

futurice-chart/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2015, Futurice OY
```
phadej: 2015 :)

PR #697, opened 2019-05-13
Update the stack.yaml for 8.6.5

jarcane:
Updates the stack.yaml to the recent changes and version of GHC, so that HIE will work again.


jarcane:
For future notes:

`mega-repo-tool stack-yaml` still seems to work.

To verify that the config will still work with HIE you can do `stack build --only-configure .`

PR #701, opened 2019-05-22
Replace ISBN lookup with Google Books API

jarcane:
does what it says in the title: replaces the Amazon stuff with Google Books.

Old Amazon links are still retained though, and the field there is renamed "info_link", as it could contain links to both Amazon or GB depending on how old the book is in the system.


library-app/library-app.cabal
```diff
@@ -66,6 +66,7 @@ library
     , swagger2
     , text
     , time
+    , unordered-containers
     , xeno
```
neurogoo: Xeno probably is not needed anymore so it can be removed

library-app/src/Futurice/App/Library.hs
```diff
@@ -146,83 +140,38 @@ addNewCover ctx cover = do
...
-
-fetchImageFromAmazon :: Ctx -> Text -> IO LBS.ByteString
-fetchImageFromAmazon ctx url = do
+    runLogT "fetch-from-google" (ctxLogger ctx) $
+        case eitherDecode $ HTTP.responseBody response of
+            Left err -> do
+                _ <- error $ "Error parsing Google Books response: " <> show err
+                pure Nothing
+            Right book -> do
+                let info = Just $ bookToInfo book
```
neurogoo: I would remove separate function and just put the function body in here

library-app/src/Futurice/App/Library/Types/GoogleBookResponse.hs
```diff
@@ -0,0 +1,62 @@
...
+    parseJSON (String "ISBN_13") = pure ISBN13
+    parseJSON _                  = mzero
+
+data ISBN = ISBN
+    { _isbnType :: !ISBNType 
+    , _isbnIdentifier :: !Text
+    }
+
+instance FromJSON ISBN where
+    parseJSON (Object v) =
```
neurogoo: This you should be able to derive via Sopica instance

library-app/src/Futurice/App/Library/Types/GoogleBookResponse.hs
```diff
@@ -0,0 +1,62 @@
...
+    , _gbrCoverLink :: !Text
+    }
+
+makeLenses ''GoogleBookResponse
+
+instance FromJSON GoogleBookResponse where
+    parseJSON (Object o) = do
+        maybeBook <- listToMaybe <$> o .: "items"
+        let book = fromMaybe HM.empty maybeBook
+        vi <- book .: "volumeInfo"
```
neurogoo: HashMap has Monoid instance so I would just use `mempty` here and not bring this HM dependency

library-app/src/Futurice/App/Library/Types/GoogleBookResponse.hs
```diff
@@ -0,0 +1,62 @@
...
+    , _gbrPublished :: !Int
+    , _gbrBooksLink :: !Text
+    , _gbrCoverLink :: !Text
+    }
+
+makeLenses ''GoogleBookResponse
+
+instance FromJSON GoogleBookResponse where
+    parseJSON (Object o) = do
+        maybeBook <- listToMaybe <$> o .: "items"
```
neurogoo: You can use withObject function to make this little cleaner

library-app/library-app.cabal
```diff
@@ -67,7 +67,6 @@ library
     , text
     , time
     , unordered-containers
```
neurogoo: Unordered-containers should not be necessary anymore as you don't use empty function from it

library-app/src/Futurice/App/Library.hs
```diff
@@ -155,20 +155,16 @@ fetchBookInformationFromGoogle ctx isbn = do
                 _ <- error $ "Error parsing Google Books response: " <> show err
                 pure Nothing
             Right book -> do
-                let info = Just $ bookToInfo book
+                let info = Just $ BookInformationByISBNResponse
```
neurogoo: Looks in my eyes cleaner if you call pure in here instead of having variable

library-app/src/Futurice/App/Library/Types/GoogleBookResponse.hs
```diff
@@ -28,11 +27,9 @@ data ISBN = ISBN
...
     }
 
-instance FromJSON ISBN where
-    parseJSON (Object v) = 
-        ISBN <$> v .: "type"
-             <*> v .: "identifier"
-    parseJSON _ = mzero
+deriveGeneric ''ISBN
+
+deriveVia [t| FromJSON ISBN `Via` Sopica ISBN |]
```
neurogoo: Prefer that we use the DerivingVia extension directly always when we have chance instead of using the TemplateHaskell invocation (some type trickery don't work with the extension, but basic cases are fine)

PR #705, opened 2019-05-28
Truncate library info link

jarcane:
it turns out Amazon links are really long and break the UI

### DIFF
--- a/library-app/src/Futurice/App/Library/Pages/BookInformationPage.hs
+++ b/library-app/src/Futurice/App/Library/Pages/BookInformationPage.hs
@@ -30,7 +30,7 @@ bookInformationPage (BookInformationResponse binfoid title isbn author publisher
                 vertRow_ "Publisher" $ toHtml $ publisher
                 vertRow_ "Published" $ toHtml $ show published
                 vertRow_ "ISBN" $ toHtml $ isbn
-                vertRow_ "Info link" $ a_ [ src_ infoLink ] $ toHtml $ T.take 80 infoLink <> "..."
+                vertRow_ "Info link" $ a_ [ src_ infoLink ] $ toHtml $ truncLink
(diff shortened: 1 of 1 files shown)


library-app/src/Futurice/App/Library/Pages/BookInformationPage.hs
```diff
@@ -30,7 +30,7 @@ bookInformationPage (BookInformationResponse binfoid title isbn author publisher
                 vertRow_ "Publisher" $ toHtml $ publisher
                 vertRow_ "Published" $ toHtml $ show published
                 vertRow_ "ISBN" $ toHtml $ isbn
-                vertRow_ "Info link" $ a_ [ src_ infoLink ] $ toHtml $ infoLink
+                vertRow_ "Info link" $ a_ [ src_ infoLink ] $ toHtml $ flip T.append "..." $ T.take 80 infoLink
```
neurogoo: Nothing wrong in this, but flip in here confused at least me a little bit. Maybe this would be a little cleaner?
```haskell
toHtml $ T.take 80 infoLink <> "...."
```

jarcane: Oh yeah, I forgot about that operator. I always get it confused because that in BASIC it means "not equals". XD

neurogoo: Also probably you should at check if the infolink is long enough to warrant "...". I guess google links can be sorter?

jarcane: oh yeah, that's a good point. I picked a number that's short enough it should almost always truncate, but people could also use shorter links even than that maybe if they don't use ISBN search to generate the Google Books link.

PR #712, opened 2019-06-28
Optimize performance of OKR report

jarcane:


jarcane:
Going to merge this now, so it can be deployed and at least the performance will be a bit better. There's still optimizations that could be done but they'll need modifications to Planmill client and I'm hesitant to risk redeploying that solo.

PR #723, opened 2019-08-30
Feature/change token policy

jarcane:
the UI here is a bit rough

experimenting with maybe making it a separate page instead but still working out how to even do that

### DIFF
--- a/futurice-foundation/src/Futurice/Lomake.hs
+++ b/futurice-foundation/src/Futurice/Lomake.hs
@@ -172,6 +172,9 @@ vHidden = VHidden
 vDynamic :: [(a, Text)] -> V a
 vDynamic = V Nothing
 
+vJustDynamic :: a -> [(a, Text)] -> V a
+vJustDynamic a opts = V (Just a) opts
+
 -- | 'maybe' for 'V'.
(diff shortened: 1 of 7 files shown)


neurogoo:
One option could be to overload the new token dialog to add new token to user or update it's policy if user is found in database

futurice-foundation/src/Futurice/Lomake.hs
```diff
@@ -172,6 +172,9 @@ vHidden = VHidden
 vDynamic :: [(a, Text)] -> V a
 vDynamic = V Nothing
 
+vJustDynamic :: a -> [(a, Text)] -> V a
```
neurogoo: Why did you need new field type?

jarcane: I cannot seem to provide *both* a default entry, *and* a dynamic field type. `vDynamic` causes it to default to empty, and `vJust` only works with regular enums, so there's no way to have a default dynamic entry showing otherwise. Or at least if there is I couldn't figure out the means, but this does what was needed.

jarcane:
Rewrote the token form to allow for upsert, seems to work great

PR #724, opened 2019-09-03
Modify ghc flags to test how personio-proxy behaves

neurogoo:


personio-proxy/src/Futurice/App/PersonioProxy.hs
```diff
@@ -126,7 +126,7 @@ makeCtx (Config cfg pgCfg intervalMin) lgr mgr cache mq = do
         let newMap' = IdMap.fromFoldable employees
         let allData' = P.PersonioAllData employees validations cl clRole
 
-#if __GLASGOW_HASKELL__ >= 804
+#if __GLASGOW_HASKELL__ >= 884
```
phadej: this is typo

neurogoo: Is it? This was lazy hack to disable compacting without deleting that part of code (as we still run ghc 8.6). Reason for this is that for some reason personio-proxy has been unstable in recent months and eats a lot of memory so trying to disable stuff and see if we get to stable place

phadej:
```haskell
#if __GLASGOW_HASKELL__ >= 804 && 0
```

would been more explicit :)

PR #725, opened 2019-09-11
mega-repo-tool checks for gpg2 on startup

jarcane:
Adds a check that makes sure GPG2 is available, and quits early with an error if its not.

### DIFF
--- a/mega-repo-tool/src/Futurice/App/MegaRepoTool.hs
+++ b/mega-repo-tool/src/Futurice/App/MegaRepoTool.hs
@@ -215,7 +215,7 @@ defaultMain :: IO ()
 defaultMain = do
     (_, _, _, testGPG) <- createProcess (shell "gpg --version") { std_out = CreatePipe }
     gpgExitCode <- waitForProcess testGPG
-    case (gpgExitCode) of
+    case gpgExitCode of
         ExitSuccess -> do
             cmd <- O.execParser opts
             main' cmd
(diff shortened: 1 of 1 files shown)


mega-repo-tool/src/Futurice/App/MegaRepoTool.hs
```diff
@@ -210,7 +212,15 @@ main' (CmdLambda flib payload mlimit) = cmdLambda flib payload mlimit
 main' (CmdCopyArtifacts bl rd bd)     = cmdCopyArtifacts (CAO rd bd bl)
 
 defaultMain :: IO ()
-defaultMain = O.execParser opts >>= main'
+defaultMain = do
+    (_, _, _, testGPG) <- createProcess (shell "gpg --version") { std_out = CreatePipe }
+    gpgExitCode <- waitForProcess testGPG
+    if (gpgExitCode == ExitSuccess) then
```
neurogoo: I would you case expression instead of if if dealing with something else than purely boolean values. Also in here, you could add the ExitFailure message to the error message to give more information to the user

mega-repo-tool/src/Futurice/App/MegaRepoTool.hs
```diff
@@ -210,7 +212,16 @@ main' (CmdLambda flib payload mlimit) = cmdLambda flib payload mlimit
 main' (CmdCopyArtifacts bl rd bd)     = cmdCopyArtifacts (CAO rd bd bl)
 
 defaultMain :: IO ()
-defaultMain = O.execParser opts >>= main'
+defaultMain = do
+    (_, _, _, testGPG) <- createProcess (shell "gpg --version") { std_out = CreatePipe }
+    gpgExitCode <- waitForProcess testGPG
+    case (gpgExitCode) of
```
neurogoo: Remove these extra brackets :) But otherwise looks good

PR #753, opened 2019-11-19
Try running in Travis with ghc-8.8.1

neurogoo:


.travis.yml
```diff
@@ -92,3 +92,14 @@ matrix:
...
         - libgmp-dev
+  - env: BUILD=cabal GHCVER=8.8.1 # PEDANTIC=YES
+    compiler: ghc-8.8.1
+    addons:
+      apt:
+        sources:
+        - hvr-ghc
+        packages:
+        - libfftw3-dev
+        - cabal-install-head
```
phadej: use `-3.0`

phadej: ... and I'd suggest starting with `futurice-haskell-prelude`, that's the first thing which would need help.

neurogoo: Heh true, forgot cabal update already :)

neurogoo:
Still need to either vendor/wait/replace few packages before continuing work:
- Haxl
- algebraic-graphs (there is already patch in Github)
- pretty-compact
- machines-* (low activity on those in recent years so may need to be replaced)

neurogoo:
Merged needed changes needed to support ghc-8.8 from another branch

PR #754, opened 2019-12-01
Remove Flowdock data check as that is not used anymore

neurogoo:


phadej:
Sunday is the new Monday?

neurogoo:
I wanted to get this done before Monday reporting cycle ^^' Not ideal, but I have been on sick leave for a while

PR #756, opened 2019-12-02
Try to parse publishedDate in multiple formats

neurogoo:


phadej:
fyi, `read` throws exceptions on invalid input. Use `readMaybe`.

neurogoo:
Yeah, I was just lazy and thinking any fail is fail for Parsing anyway, but true, I should change that to more proper handling

PR #765, opened 2020-01-08
Add PersonioId to Value Creation report

mixman:
Convert PlanMill's "FirstName,LastName" identifier to PersonioId.

### DIFF
--- a/reports-app/src/Futurice/App/Reports/ValueCreation.hs
+++ b/reports-app/src/Futurice/App/Reports/ValueCreation.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE TemplateHaskell   #-}
 module Futurice.App.Reports.ValueCreation where
 
 import Futurice.Integrations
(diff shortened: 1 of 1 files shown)


reports-app/src/Futurice/App/Reports/ValueCreation.hs
```diff
@@ -7,8 +9,23 @@ import Futurice.Prelude
 import Prelude ()
 import Futurice.Time.Month (monthYear)
 
+import qualified Data.HashMap.Strict as HM
 import qualified PlanMill         as PM
+import qualified Personio           as P
 import qualified PlanMill.Queries as PMQ
+import qualified Data.Text           as T
+
+import qualified Personio.Types.Status as PS
```
neurogoo: Personio does export Status already so there should be no need to import it separately

reports-app/src/Futurice/App/Reports/ValueCreation.hs
```diff
@@ -7,8 +9,23 @@ import Futurice.Prelude
...
+
+-- Create a mapping for matching PlanMill names to PersonioID's
+planmillNameToPersonioIdMap
+    :: (MonadPersonio m)
+    => m (HashMap (Text, Text) P.Employee)
+planmillNameToPersonioIdMap = do
+  us <- P.personio P.PersonioEmployees
+  let usActive = filter (\p -> p ^. P.employeeStatus == PS.Active) us
+  let us' = map (\p -> ((p ^. P.employeeFirst, p ^.  P.employeeLast), p)) usActive
+  pure $ HM.fromList us'
```
neurogoo: Why HashMap instead of regular Map? Not that it effect much to this, but I would usually use Map unless there is reason for otherwise

reports-app/src/Futurice/App/Reports/ValueCreation.hs
```diff
@@ -36,15 +54,22 @@ data ValueCreationRow = ValueCreationRow
...
+    pure $ ValueCreationReport year $ fmap (\x -> toRow nameMap x) $ toList $ PM.getValueCreationData report
   where
-    toRow v = ValueCreationRow
+    nameSplit v = case (T.splitOn "," $ PM._pvcPerson v) of
+                    (f:l:_) -> (f,l)
+                    _       -> ("", "")
+    toRow nm v = ValueCreationRow
         { valueRowTeam      = PM._pvcTeam v
         , valueRowPerson    = PM._pvcPerson v
+        , valueRowPersonioId = case HM.lookup (nameSplit v) nm of
```
neurogoo: instead of returning ("",""), I would return Nothing and then write this part as `nameSplit v >>= flip HM.lookup nm`

reports-app/src/Futurice/App/Reports/ValueCreation.hs
mixman: No real reason. I saw both are used within the codebase and hadn't tried HashMap before.

PR #767, opened 2020-01-27
Gather paged value-creation data for a full report

mixman:
Prior the report was limited to default 1000 entries.

### DIFF
--- a/planmill-client/src/Control/Monad/PlanMill.hs
+++ b/planmill-client/src/Control/Monad/PlanMill.hs
@@ -68,7 +68,6 @@ class
     , MonadPlanMillC m UserCapacity
     , ForallFSymbol (MonadPlanMillC m) EnumDesc
     , MonadPlanMillC m AllRevenues2
-    , MonadPlanMillC m ValueCreationByMonth
     , MonadPlanMillC m PersonValueCreation
     )
   => MonadPlanMillConstraint m where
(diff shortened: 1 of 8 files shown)


planmill-client/src/Control/Monad/PlanMill.hs
```diff
@@ -69,6 +69,7 @@ class
     , ForallFSymbol (MonadPlanMillC m) EnumDesc
     , MonadPlanMillC m AllRevenues2
     , MonadPlanMillC m ValueCreationByMonth
```
neurogoo: I assume this could be removed as now it is just type alias

planmill-client/src/PlanMill/Types/Report.hs
```diff
@@ -154,6 +145,7 @@ data PersonValueCreation = PersonValueCreation
     , _pvcNovemberValue  :: !Double
     , _pvcDecemberValue  :: !Double
     }  deriving (Eq, Show, Binary, GhcGeneric, SopGeneric, ToSchema, ToJSON, NFData, HasSemanticVersion, HasDatatypeInfo)
+type ValueCreationByMonth = Vector PersonValueCreation
```
neurogoo: I would change this to be `type PersonValueCreations = Vector PersonValueCreation` to make it even more clear what is the relationship of the types

mixman: Sure. The type name confused me too.

PR #778, opened 2020-02-24
Add power tribe monthly forecast report to prox

mixman:


proxy-types/src/Futurice/App/Proxy/API.hs
```diff
@@ -132,6 +132,10 @@ data Routes = Routes
         ("uiapi" :> "tribe/" :> Get '[JSON] Value)
         ("power" :> "api" :> "v3" :> "tribes" :> Get '[JSON] Value)
 
+    , routePowerV3TribesMonthlyForecast :: ProxiedEndpoint 'PowerService
+        ("uiapi" :> "tribe" :> "monthly_forecast/" :> Get '[JSON] Value)
+        ("power" :> "api" :> "v3" :> "tribes" :> "monthly" :> "forecast" :> Get '[JSON] Value)
```
neurogoo: For us, it is a little easier if the endpoints are similar than in source. Or is there reason to change that?

mixman: What change are you looking for?

neurogoo: both endpoints to be either "montly_forecast" or "monthly/forecast"

mixman: Ah! Too late for Power without refactoring there.

mixman: Settled offline. Will migrate in future.

PR #780, opened 2020-03-10
proxy allocation history

mixman:


proxy-types/src/Futurice/App/Proxy/API.hs
```diff
@@ -140,6 +140,10 @@ data Routes = Routes
         ("uiapi" :> "allocation" :> "" :> QueryParam "start_date" Day :> QueryParam "end_date" Day :> Get '[JSON] Value)
         ("power" :> "api" :> "v3" :> "allocations" :> QueryParam "start_date" Day :> QueryParam "end_date" Day :> Get '[JSON] Value)
 
+    , routePowerV3AllocationsHistory :: ProxiedEndpoint 'PowerService
+        ("uiapi" :> "allocation" :> "history/" :> Get '[JSON] Value)
```
neurogoo: instead of explicit '/' in the end I would write it like `"uiapi" :> "allocation" :> "history" :> "" :> Get '[JSON] Value`

neurogoo: There is some encoding problems when writing explicit '/' like that

mixman: Sure, as long as "" is the same as /? With Power the ending slash is mandatory.

Encoding problems with aeson or servant or what?

neurogoo: Should be the same. I don't remember the exact problem, but basically servant treats the '/' differently depending where it is

PR #838, opened 2020-10-16
Add another field to the report for Inari

neurogoo:


phadej:
Opening an office in Lapland? :)

neurogoo:
> Opening an office in Lapland? :)

Heh, not exactly 😄 Name of internal data platform

PR #886, opened 2022-01-27
Fill personio-proxy frontend profiles with good data

lumicake:
I implemented some logic that kind of "casts" the `P.Employee` data to `EmployeeSafe` and leverages `Maybe` and Aeson's `omitNothingFields` via `toJSON`. I played around with the idea of creating a `ToJSON` instance for `P.Employee` that would just build the JSON object a specific way, but it already has a `ToJSON` instance defined in Employee Types, and apparently there can only be one :sweat_smile: I don't know if this is the best way to solve this problem but it is a way to solve it :D


personio-proxy/src/Futurice/App/PersonioProxy/IndexPage.hs
```diff
@@ -145,6 +148,71 @@ indexPage ps = page_ "Personio Proxy" (Just NavHome) $ do
                 td_ $ toHtml $ e ^. P.employeeStatus
                 td_ $ traverse_ toHtml $ e ^. P.employeeEmploymentType
 
+makeEmployeeSafe :: P.Employee -> EmployeeSafe
+makeEmployeeSafe p = EmployeeSafe
+    (p ^. P.employeeFirst)
```
mixman, line 153: The convention, to avoid positional mistakes, is to use full record syntax `{ _first = p ^. P.employeeFirst , ...`

I know, it's a lot of typing...

mixman (approved):
LGTM. Similar effort as other Employee records in `Personio/Types`.

personio-proxy/src/Futurice/App/PersonioProxy/IndexPage.hs
lumicake: Fixed!

PR #887, opened 2022-01-27
Refactor/adds new company to hmr

thaiscristine:


futurice-tribes/src/Futurice/Company.hs
```diff
@@ -88,6 +89,9 @@ companyFuturiceGmbH = fromMaybe (error "Company Futurice GmbH") $ companyFromTex
 companyFutuSwedenAB :: Company
 companyFutuSwedenAB = fromMaybe (error "Company FutuSweden AB") $ companyFromText "Futu Sweden AB"
 
+companyFuturiceHubsOy :: Company
```
thaiscristine, line 92: @mixman Are these lines needed at all? Or only the .json file from hmrs is enough to add new company data?

mixman: JSON should be enough. I didn't pinpoint immediate use-case. Some of these could be related to some UI component / WIP.

lumicake:
You will need to update the sha256 sum for `data/tribes.json` if you add the new tribe/department that was specified for Hubs.

thaiscristine:
> You will need to update the sha256 sum for `data/tribes.json` if you add the new tribe/department that was specified for Hubs.

Good catch. Updated 👌🏽

JoshStrobl (approved):
LGTM, thanks!

lumicake (approved):
LGTM :rocket:

PR #888, opened 2022-02-18
Fix/syncing issues fixes

lumicake:
These changes are relatively self-explanatory.
1) The variable salary dynamic field whose validation is disabled in this commit is apparently not in use by HC anymore. Should actually just probs delete the code instead of commenting it out.
2) Added the "Germany remote" office so that it does not keep breaking our personio data sync from time to time. The actual addition is in the secrets repo, here it's just the checksum that has been updated.
3) Made `cabal` aware of the `data/*.json` files so that it can track them and much smaller builds suffice when changes to them are made.

## Related PR
futurice/haskell-mega-repo-secrets#9


JoshStrobl (approved):
LGTM, thanks!

mixman (approved):
LGTM, good notes for future

thaiscristine (approved):
LGTM. Good job adding the comments!

PR #890, opened 2022-02-23
Move from Travis CI to GitHub Actions

JoshStrobl:
Seems most of the Futurice repos are moving away from Travis CI and run into various billing-related issues. GitHub Actions is perfectly suitable and works better for our GitHub-based workflow, so let's swap over to that.

Special thanks to @jsavikko for the massive amount of foundational work on this. Much appreciated.


.github/workflows/haskell.yml
```diff
@@ -0,0 +1,95 @@
...
+      - github-actions
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    container: ubuntu:latest
+    strategy:
+      matrix:
+        ghc: ['latest']
```
lumicake, line 19: Are we sure we don't want to constrain the version of `ghc`? I have seen some commits in the history where actual changes to the codebase have had to be made to get everything to work with newer `ghc` versions. And as I recall, @mixman spent a bunch of time on similar changes to get us to 8.10.7 I believe? #877

lumicake (approved):
LGTM :100:  Thanks for figuring the CI situation out :) I had one comment about constraining the `ghc` version, might be a smart idea to do that, or maybe not :woman_shrugging:

It looked like the old travis solution used `ghcup` for something -- or did it? Kinda looked like maybe it was just installed and not used for anything, weird.

.github/workflows/haskell.yml
JoshStrobl: Thanks for pointing that out. I believe @jsavikko  was having issues with ghlint and ghc 8.10.x. Will take a poke at that again!

jsavikko: I was following the recommendation of the [haskell/actions/setup](https://github.com/haskell/actions/tree/main/setup) when defining the version.

At first I had defined two versions ['8.8.4', '8.10.7'] based on the [Travis configuration](https://github.com/futurice/haskell-mega-repo/blob/master/.travis.yml#L43-L49), but the setup had some problems with ghc version [8.8.4](https://github.com/futurice/haskell-mega-repo/runs/5299682869?check_suite_focus=true). Considering that 8.8.4 doesn't have support for Apple M1 ([support came in 8.10.5](https://www.haskell.org/ghc/blog/20210309-apple-m1-story.html)), I guess we can drop at least that from the list.

I think no one has actually used too much time to look at those version things. We had a person in the team who was involved in GHC development and was building the compiler and packages on our infrastructure. But I would say that since then Haskell has not been too much in focus in our team or systems.

So, my opinion is that we can build things on both 8.10.7 and latest. I just defined the latest to have faster "github actions" development testing cycles :)

lumicake:
Yep, sounds good to me! I wonder why `8.8.4` was left in there in the first
place when we updated to `8.10.7`. Now is a good time to get rid of it
though :)

.github/workflows/haskell.yml
lumicake: Email replying put my reply in a bit of a weird place :D But:

Yep, sounds good to me! I wonder why `8.8.4` was left in there in the first
place when we updated to `8.10.7`. Now is a good time to get rid of it
though :)

mixman: 8.10.7 is the latest version the codebase runs on while being backwards compatible with 8.8.4. Newer versions not tested.

PR #891, opened 2022-02-24
Add planmillId to power/users report

mixman:
Allows mapping data between Power<>Personio<>PlanMill within Power.

There was some issues outputting mega-repo-tool config, fixed by changing printer.


mixman:
Looking at this again I noticed I should use the emailMap, given changes to login and desire to move away from them.

mixman:
Looking at this again I noticed I should use the emailMap, given changes to login and desire to move away from them.

PR #893, opened 2022-03-17
Add new field to be synced from Personio to Okta

lumicake:
IT support requested us to add a new Personio field to be synced to Okta. I believe these changes are sufficient, but I have no way of testing these changes. I have confirmed that these changes do fetch the data from Personio correctly and put them in the db of `personio-proxy`, but I don't think I can test the actual Okta side of things at all.

This is currently pending some answers from IT support: I am unsure of whether some setup is required on the Okta side of things so that Okta knows to expect this data.


mixman (approved):
Looks fine. Given Okta profile has the field with that name it should work. Good to sync with IT to ensure all OK.

JoshStrobl (approved):
LGTM!

PR #902, opened 2022-06-29
Chore/add impact roles okta sync

thaiscristine:


thaiscristine:
This PR is to solve [this ticket](https://support.futurice.com/helpdesk/tickets/26175). @lumicake Do you know if any extra step is required? Such as updating hmrs, for example

jvanbruegge (approved):

personio-client/src/Personio/Types/Employee.hs
```diff
@@ -93,6 +93,7 @@ data Employee = Employee
     , _employeeNationality       :: !(Maybe Text)
     , _employeePublicCareerLevel :: !(Maybe CareerLevel)
     , _employeePowerBiSecurityGroup       :: !(Maybe Text)
+    , _employeeImpactRoles       :: !(Maybe Text)
```
lumicake, line 96: This data is coming from Personio? You need to add an accompanying line in the actual parsing of the data from Personio, I believe.

For example, look at line 95:
```haskell
, _employeePowerBiSecurityGroup       :: !(Maybe Text)
```

it has an accompanying line (line#214) in the `parseEmployeeObject` function:
```haskell
<*> optional (parseDynamicAttribute obj "Access rights to PowerBI security group")
```

You need to add this kind of parsing for impact roles as well.

lumicake (changes requested):
Looks good, I think there's one change needed for this to actually work.

lumicake:
> 

I believe these changes should be sufficient :)

PR #903, opened 2022-07-05
Decouple avatar from FUM

lumicake:
This is a very crude implementation of decoupling Avatar from FUM. I don't think it should be used as is, but maybe @jvanbruegge or me or someone else can use it as a basis :eyes:


lumicake:
Also these changes need to be rebased onto the FUM sunset changes, but when I was making these changes I tried to take that into account to make it as easy as possible.

PR #906, opened 2022-09-09
Prevent people who have not accepted job offer from being in reports

lumicake:
A kinda dirty fix, open to ideas for improvement. Ideally this kind of check could be implemented where the `p<personioId>` injection happens in `personio-client`, but I don't feel comfortable touching that code. A better fix needs to be applied later, there are some ideas that need to be investigated.


lumicake:
This actually only prevents these people from showing up in Power. This fix should probs be applied to some other places in `reports-app` too and maybe `contacts`?

mixman:
Yes, this report is only for Power.

